PR #21, opened 2018-08-01
Merge config

harshurampur:
Update the prod services for geomedian-au to ``` DEA geomedian terrain corrected surface reflectance```

Add wofls and wofs_summary to wofs layer

### DIFF
--- a/dev/services/wms/ows/wms_cfg.py
+++ b/dev/services/wms/ows/wms_cfg.py
@@ -1020,19 +1020,41 @@ layer_cfg = [
                     {
                         "name": "mangrove",
                         "title": "Mangrove Cover",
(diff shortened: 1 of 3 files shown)


dev/services/wms/ows/wms_cfg.py
```diff
@@ -988,12 +997,58 @@
             }
         ]
     },
+    {
```
roarmstrong: Please use the mangrove styles found here: https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/mangrove/wms_cfg.py

harshurampur: updated with mangroves

PR #33, opened 2018-09-03
Add LS5_fc product definition

GypsyBojangles:
Add munged LS5_fc product definition - should be consistent with the Landsat 5 Fractional cover that has been uploaded to AWS. (untested)

### DIFF
--- a/dev/products/fc/ls5_fc_albers.yaml
+++ b/dev/products/fc/ls5_fc_albers.yaml
@@ -1,0 +1,42 @@
+name: ls5_fc_albers
+
+description: Landsat 5 Fractional Cover 25 metre, 100km tile, Australian Albers Equal Area projection (EPSG:3577)
+metadata_type: eo
+
+metadata:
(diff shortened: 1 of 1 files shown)


dev/products/fc/ls5_fc_albers.yaml
```diff
@@ -0,0 +1,42 @@
+name: ls5_fc_albers
+
+description: Landsat 5 Fractional Cover 25 metre, 100km tile, Australian Albers Equal Area projection (EPSG:3577)
+metadata_type: eo
+
+metadata:
+  platform:
+    code: LANDSAT_5
+  format:
+    name: GeoTIFF
```
omad: :( :( :( ;(

I think we need to stop putting `format: name: GeoTIFF/NetCDF` into our product specs. AFAICT it's completely unnecessary, and means we're duplicating the definitions all over the place to change one value. ie. maintenance nightmare

PR #97, opened 2018-10-10
update color-ramp values to jet colormap available im matplotlib

harshurampur:


omad:
Why jet? It's been replaced by so many better options!

https://matplotlib.org/users/colormaps.html
https://matplotlib.org/cmocean/

harshurampur:
Jet mapping was suggested from Leo to all the styles

omad:
> Jet mapping was suggested from Leo to all the styles

I'll take it up with him!

PR #133, opened 2018-10-17
add product config for nidem

harshurampur:


omad:
Hi Harshu,

I think `instrument` and `platform` should be removed. NIDEM isn't an 'observation' from a remote sensing platform, it's an output from an algorithm.

omad:
Thanks

Also, it's stored in EPSG:3577, not 4326.

@robbibt Which bands do we want for NIDEM in Data Cube? Which bands do we want uploaded to S3 and available through the web services?

<details>
<summary>As an example, here's what the ITEM product looks like</summary>

```yaml
description: Intertidal Extents Model
managed: true
measurements:
-   dtype: int16
    name: relative
    nodata: 0
    units: '0.1'
metadata:
    format:
        name: NetCDF
    product_type: ITEM
    statistic:
        name: decile
metadata_type: eo
name: item_v2
storage:
    crs: EPSG:3577
    resolution:
        x: 25
        y: -25
```
</summary>

I don't actually think we should be using the `eo` `metadata_type`, but lets ignore that for now and just get something working.

harshurampur:
Yesterday Robbi mentioned just to use `nidem` variables

dev/products/nidem/nidem.yaml
```diff
@@ -0,0 +1,18 @@
+name: nidem
+description: National Intertidal Digital Elevation Model 25m 1.0.0
+metadata_type: eo
+
+metadata:
+    format:
+        name: GeoTIFF
+    product_type: item_v2.0
```
omad:
```suggestion
    product_type: nidem_v1.0.0
```

dev/products/nidem/nidem.yaml
```diff
@@ -0,0 +1,18 @@
...
+measurements:
+-   dtype: float32
+    name: 'nidem'
+    nodata: -9999
+    units: '1'
```
omad:
```suggestion
    units: 'metres'
```
NIDEM is an elevation in

PR #152, opened 2018-10-23
Production NDVI, NDWI color ramp changes & Geomedian single band

roarmstrong:
Update production ndvi and ndwi styles for new legend configuration. Add single band styles to geomedian products


sixy6e:
If this is going into production I'd suggest using the FWHM instead of "centred".  Just stating the wavelength centre is just as uninformative as a name.

roarmstrong:
Thanks for the suggestion @sixy6e! Will raise it with the product owners and see if they are keen to use a more descriptive name

PR #174, opened 2018-10-28
modify legends for ITEM relative layer

harshurampur:


dev/services/wms/ows/wms_cfg.py
```diff
@@ -4494,72 +4494,50 @@
...
-                                'alpha': 1.0,
                                 'legend': {
                                     "label": "70-80% and above"
                                 }
                             },
                         ],
+                        "legend": {
+                            "units": "%",
+                            "radix_point": 0,
+                            "scale_by": 100.0,
```
roarmstrong: Should scale by `10.0` to get a percentage displayed in the legend

roarmstrong:
```suggestion
                            "scale_by": 10.0,
```

PR #199, opened 2018-11-26
Fix the barest_earth_mosaic tile size

ASVincent:
- Jeremy identified an issue with the barest earth mosaic tile size in the product config. This change propagates that fix to the dea-config repo and our web services. 

Related Issue: https://github.com/GeoscienceAustralia/digitalearthau/issues/190

Related PR: https://github.com/GeoscienceAustralia/digitalearthau/pull/191

Looks like it doesn't affect the other dev config or prod config.

### DIFF
--- a/dev/products/ls8-bare-earth/landsat8_barest_earth_mosaic.yaml
+++ b/dev/products/ls8-bare-earth/landsat8_barest_earth_mosaic.yaml
@@ -13,8 +13,8 @@ storage:
   driver: GeoTIFF
   crs: EPSG:3577
   tile_size:
-    x: 25000.0
-    y: 25000.0
+    x: 100000.0
+    y: 100000.0
   resolution:
     x: 25
     y: -25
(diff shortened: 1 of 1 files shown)


jeremyh:
Thanks @harshurampur 

So the services are only using the `_albers` product definition, but the NCI instance is only using the [`_mosaic` definition](https://data.dea.gadevs.ga/ls8_barest_earth_mosaic) (and has 100km datasets).

Should we switch NCI to use the same `_albers` product definition, or are they different datasets?

harshurampur:
@jeremyh 
When the tiles were generated for Barest Earth they were 4X4 of each Albers size tile. Andrew ingested them to be Albers specific as 100kmx100km. So there are copies of 25km and 100km and named mosaic to be 25km and landsat8_barest_earth_albers to be 100km. If we switch the name to be `Albers` for 100kmx100km then will minimize the confusion

PR #214, opened 2018-12-11
remove wet legend, as it is in the original legend

tom-butler:


dev/services/wms/ows/wms_cfg.py
```diff
@@ -2577,7 +2577,7 @@
                                         "or": {
                                           "terrain_or_low_angle": True,
                                           "cloud_shadow": True,
-                                          "cloud": True,
+                                          "cloud": True,gi
```
roarmstrong:
```suggestion
                                          "cloud": True,
```

PR #251, opened 2019-01-31
Remove threshold of NDCI to make larger water bodies look smoother

tom-butler:


robbibt:
@tom-butler Just noticed that these changes only apply to `s2a`, not `s2b`.

These lines need to be updated too:

https://github.com/GeoscienceAustralia/dea-config/blob/82acdb432ba04f1289b5afc62af55f3327f090a5/dev/services/wms/ows/wms_cfg.py#L3742

https://github.com/GeoscienceAustralia/dea-config/blob/82acdb432ba04f1289b5afc62af55f3327f090a5/dev/services/wms/ows/wms_cfg.py#L3748

PR #283, opened 2019-02-28
Add ASTER Layers to Terria

GypsyBojangles:
Add ASTER Layers to terria for testing by Resources


GypsyBojangles:
It looks like i made a mistake with the layer names

PR #288, opened 2019-03-06
Updating definitions for Sentinel-2 products to handle SAFE-l1c derived products

ASVincent:
Proposed changes to allow uploading Sentinel-2 definitive products for populating the 90 backlog of the NRT service.

* Removes format: name (file format) from being a differentiating field for Sentinel-2 products
* Adds product_type as the main discriminator between aws pds and SENTINEL-SAFE level 1 formats
* Adds a product definition for the SENTINEL-2 SAFE product (in lineage)


tom-butler:
No Longer required

PR #294, opened 2019-03-17
Merge FC products into single group. Remove pq_datasets

roarmstrong:


dev/services/wms/ows/wms_cfg.py
```diff
@@ -10975,37 +10975,42 @@
...
 Fractional Cover products use Water Observations from Space (WOfS) to mask out areas of water, cloud and other phenomena. To be considered in the FCP product a pixel must have had at least 10 clear observations over the year.
 
-This product is dervied from the Landsat 5 satellite.
-
 For service status information, see https://status.dea.ga.gov.au""",
         "products": [
             {
             # Included as a keyword  for the layer
-            "label": "FC",
+            "label": "FCLS5",
```
harshurampur: It's best if we expand it

PR #317, opened 2019-04-04
 add new product definition for Sentinel2 WOfS

harshurampur:


dev/products/nrt/wofs/sentinel2_wofs.yaml
```diff
@@ -0,0 +1,54 @@
+name: sentinel2_wofs_albers
+description: Sentinel 2 NRT Water Observations from Space
+metadata_type: eo
+metadata:
+  product_type: wofs
+  platform:
+    code: SENTINEL_2B
```
roarmstrong: Is it only a single platform?

PR #345, opened 2019-05-17
Add barest earth 30y config in ows/wms_cfg

santoshamohan:
Add barest earth 30y config in ows/wms_cfg

### DIFF
--- a/dev/services/wms/ows/wms_cfg.py
+++ b/dev/services/wms/ows/wms_cfg.py
@@ -1357,6 +1357,297 @@ For service status information, see https://status.dea.ga.gov.au""",
         ]
     },
     {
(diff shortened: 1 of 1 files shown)


dev/services/wms/ows/wms_cfg.py
```diff
@@ -1360,16 +1360,16 @@
         # Name and title of the platform layer.
         # Platform layers are not mappable. The name is for internal server use only.
         "name": "landsat_barest_earth",
-        "title": "Barest Earth 30 Years",
+        "title": "Barest Earth 30 Years (100km x 100km tile)",
```
harshurampur: You don't need to specify the tile grid if it is in albers

PR #390, opened 2019-08-16
Add ecat pids for aster derived products

santoshamohan:


dev/services/wms/ows/wms_cfg.py
```diff
@@ -9863,6 +9867,15 @@
                 # Included as a keyword for the layer
                 "variant": "",
                 "abstract": """
+This is the parent datafile of a dataset that comprises a set of 14+ geoscience products made up of mosaiced ASTER (Advanced Spaceborne Thermal Emission and Reflection Radiometer) scenes across Australia.
```
harshurampur: This looks good!.
Just to look nicer can you please add at the beginning  as `The National ASTER Map of Australia` is... on all sub products

santoshamohan: Updated the abstract description

PR #391, opened 2019-08-19
Add aster products to prod wms config

santoshamohan:
Copy `aster` product and its derived products from `dev` `wms config` file to `prod `wms config` file


santoshamohan:
Removed unused `prod` `wms` config file

santoshamohan:
Stale `PR`. Retaining `prod/../wms_cfg.py` config file for future use.

PR #411, opened 2019-10-10
Update chart presentation

rowanwins:
This PR updates the styling of charts so that availability of imagery is now plotted as dots along a line.

Don't merge until TerriaCube has been updated to the latest version of TerriaJS.


rowanwins:
Also moves one of the GSKY services from HTTP to HTTPS

PR #420, opened 2019-10-28
Add waterbody feature template to prod

tom-butler:


tom-butler:
This was missing from the original change, and meant the wamm results weren't being graphed at all.

PR #438, opened 2019-11-20
Remove WAMM layers

rowanwins:
As per request remove WAMM layers.


tom-butler:
This seems a bit odd, just confirming the reason for this change internally before we merge it.

PR #459, opened 2019-11-29
Adding of approved text to WAMM wms.

vnewey:
We are temporarily bringing this back up for a demo

### DIFF
--- a/dev/services/wms/ows/wms_cfg.py
+++ b/dev/services/wms/ows/wms_cfg.py
@@ -9774,89 +9774,13 @@ For service status information, see https://status.dea.ga.gov.au""",
     {
         # Name and title of the platform layer.
         # Platform layers are not mappable. The name is for internal server use only.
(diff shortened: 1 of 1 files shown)


dev/services/wms/ows/wms_cfg.py
```diff
@@ -9774,13 +9774,89 @@
     {
         # Name and title of the platform layer.
         # Platform layers are not mappable. The name is for internal server use only.
-        "name": "Historical Airborne Photography",
```
GypsyBojangles: maybe we should seperate HAP from WAMM better?

GypsyBojangles: it would make adding and removing WAMM slightly easier...

vnewey: good idea!
I'm not sure why they are together. But I was try to do as minimal change as possible.

Issue #470, opened 2019-12-16
Deploy INSAR Derived product


whatnick:
Deploy InSAR derived displacement and velocity over NSW.

PR #471, opened 2019-12-17
#470 Add Terria config

whatnick:
Deploy layer for terria via ows-config refactor server.


whatnick:
> dev config-refactor won't exist for very long, can we point this to ows.dev.dea.ga.gov.au instead? and just merge it when config-refactor has been deployed.

Pointed to ows.dev

PR #489, opened 2020-01-09
Remove ows-config from CORS list

pindge:


pindge:
@tom-butler is there any other link on the list that no long exists?

PR #494, opened 2020-01-15
config files for FC alchemist and wofs alchemist

dunkgray:
Let's get this onto the main branch.


dunkgray:
Ha ha!

________________________________

PR #504, opened 2020-02-11
Add Production combined S2 NRT layer

GypsyBojangles:
Add Production combined S2 NRT layer to Terria catalogue.


whatnick:
I am uncomfortable with lack of JSLINT for JSON configs in this repo. @GypsyBojangles raising an issue to address in the future.

PR #507, opened 2020-02-21
Update wofs-addqueue.yaml

dunkgray:


dunkgray:
Thanks for merging these comits Alex

PR #519, opened 2020-04-06
Update annual to format Year: YYYY

pindge:


dev/terria/dea.json
```diff
@@ -140,7 +140,11 @@
                            "ignoreUnknownTileErrors": true,
                            "chartType": "momentPoints",
                            "chartColor": "white",
-                           "opacity": 1
+                           "opacity": 1,
+                           "dateFormat": {
+                              "currentTime": "'Year: 'yyyy",
```
tom-butler, line 145: what's going on with the ' ?

pindge: https://github.com/TerriaJS/aremi-natmap/blob/master/datasources/aremi/custodian/Institute_for_Sustainable_Futures/DANCE3/aremi-isf/constraints/ejs/constraints.ejs#L25-L28

this is actually the format
```
  "dateFormat": {
    "currentTime": "'Year: 'yyyy",
    "timelineTic": "yyyy"
  },
```

PR #520, opened 2020-04-08
Moving HAP product to folder

GypsyBojangles:
As per convention, moving Historical Airborne Photography proof of concept product definition to it's own folder


alexgleith:
Pretty risky change... I think we need some CCB approvals!

PR #543, opened 2020-06-03
Nbr

pindge:


pindge:
image

legend is compatible with https://github.com/opendatacube/datacube-ows/pull/290

PR #550, opened 2020-06-18
Add s2_gm_seasonal product config

GypsyBojangles:
Add missing s2_gm_seasonal product config


GypsyBojangles:
@alexgleith has verified that this was working (via product add):
```
datacube product add s2_gm_seasonal.yaml 
Adding "s2_gm_seasonal" (this might take a while) DONE
```

PR #572, opened 2020-09-02
New s2 nrt metadata and product definitions

alexgleith:
Should fix issues with dataset metadata needing to be patched.


whatnick:
We can swap this over, but should update manifests in CSV's to reflect.

uchchwhash (approved):
LGTM! :+1:

PR #573, opened 2020-09-02
Revert "Manual merge is broken"

pindge:
Reverts GeoscienceAustralia/dea-config#570

datacube-ows will fix manual merge


uchchwhash (approved):
If you say so.

PR #587, opened 2020-09-22
rename s2 nrt products

pindge:
following the discussion from sprint planning, s2 nrt products will continue to use its current name.


alexgleith (approved):
We're NOT changing the name, right? In that case, I approve.

PR #594, opened 2020-09-29
Add level1 product to level2 product definition

pindge:
Running indexing without level1 product, must pass in `--confirm-ignore-lineage`, for completeness it should be in the same file as level2 definition.


alexgleith:
I don't think we should be doing lineage with S-2 NRT.

Thoughts @GypsyBojangles?

PR #596, opened 2020-10-08
Added "sensitive" versions of tmad styles.

SpacemanPaul:
"Sensitive" versions of TMAD styles as requested by Dale.

Uses config inheritance.


SpacemanPaul:
The OWS bugfix required by this config patch has been merged.

This config patch should be safe to merge.

whatnick (approved):

PR #605, opened 2020-10-20
Adding in WO and FC dev config files

alexgleith:
* Add the WO config
* Add the FC config


omad:
Looks like you only did a partial rename (aka: delete). I can't see what the intended file names is.

The content looks okay. Although there's only input of LS8 for the FC product.

Are product names generated based on the provided metadata?

alexgleith:
Thanks @omad 

Yeah, not really sure how product names come about, but I think it's from metadata, yeah.

omad (approved):
LGTM 👍

PR #608, opened 2020-10-28
Patch wo config for Alchemist

alexgleith:


omad (approved):
Oh ewww okay.

I don't know what to think of yaml/alchemist with this :(

PR #612, opened 2020-10-29
Update and rename ga_ls_tcw_percentiles_2.yaml to ga_ls_tcw_percentil…

vnewey:
…es_mys_2.yaml

Change to name tof product to ga_ls_tcw_percentiles_mys_2


omad (approved):
Thanks @vnewey

alexgleith (commented):
I'd prefer we left it the way it is. Declined!

PR #630, opened 2020-12-02
Added unique titles to the TCW percentiles layers/bands.

vnewey:


pindge (approved):

dev/services/wms/ows/ows_cfg.py
```diff
@@ -327,7 +327,7 @@
 legend_tcw_400ticks = {
```
pindge, line 327: This can be deleted if we decide to use png urls

pindge (changes requested):

pindge (approved):

PR #631, opened 2020-12-02
Fix tassleed cap

pindge:


vnewey (changes requested):
Actually we are going with Tasseled.  not  Tasselled.  Because it was spelt with one el in the paper that describes the algorithm.

vnewey (approved):
Looks great!

PR #637, opened 2020-12-14
implement in ows cache

pindge:


SpacemanPaul (approved):
Looks like a good first try.    Remember to profile the original cache performance before pushing to prod.

PR #641, opened 2021-01-18
keep to c3 wo and fc only

pindge:
For paul's visibility.

### DIFF
--- a/dev/services/wms/ows/c3_ows_cfg.py
+++ b/dev/services/wms/ows/c3_ows_cfg.py
@@ -1,14 +1,5 @@
-bands_collection3_landsat = {
-    "nbart_blue": ["nbart_blue"],
-    "nbart_green": ["nbart_green"],
-    "nbart_red": ["nbart_red"],
-    "nbart_nir": ["nbart_nir", "nbart_near_infrared"],
-    "nbart_swir_1": ["nbart_swir_1", "nbart_shortwave_infrared_1"],
-    "nbart_swir_2": ["nbart_swir_2", "nbart_shortwave_infrared_2"],
-}
 
-
-reslim_landsat = {
+reslim_wofs_obs = {
     "wms": {
         "zoomed_out_fill_colour": [150, 180, 200, 160],
         "min_zoom_factor": 35.0,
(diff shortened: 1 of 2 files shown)


whatnick (changes requested):
We should aim for a single, refactored ows_cfg object. With subsets of layers created in separate python files with the existing ows_cfg aggregating them.

dev/services/wms/ows/c3_ows_cfg.py
```diff
@@ -0,0 +1,469 @@
...
+            #     "abstract": "Wet or Sea",
+            #     "flags": {"or": {"wet": True, "sea": True}},
+            #     "color": "#4F81BD",
+            # },
+        ]
+    },
+}
+
+
+ows_cfg = {
```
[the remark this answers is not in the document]
pindge: @SpacemanPaul , is now the perfect opportunity to build ows with fragmented ows_cfg files?

dev/services/wms/ows/c3_ows_cfg.py
```diff
@@ -0,0 +1,363 @@
...
+            #     "flags": {
+            #         "dry": True,
+            #         "sea": False,
+            #     },
+            #     "color": "#D99694",
+            # },
+            # {
+            #     "title": "Wet",
+            #     "abstract": "Wet or Sea",
+            #     "flags": {"or": {"wet": True, "sea": True}},
```
pindge, line 86: @GypsyBojangles I need some advice on what to do with `sea` flag

dev/services/wms/ows/c3_ows_cfg.py
```diff
@@ -0,0 +1,363 @@
...
+        {
+            "flags": {"dry": True},
+        },
+        {
+            "flags": {
+                # "terrain_or_low_angle": False,
+                "high_slope": False,
+                "cloud_shadow": False,
+                "cloud": False,
+                # "sea": False,
```
pindge, line 36: @GypsyBojangles this fc style with `pq_masks` enabled doesn't show dense green anymore, which looks very similar to fc_albers, i will need to look at the coastal areas to see how much impact it is without `sea` flag

dev/services/wms/ows/c3_ows_cfg.py
GypsyBojangles: To summarize our conversation with Team X:
- When we replace the old flag "terrain_or_low_angle", we should replace with both "low_solar_angle" and "terrain_shadow". Lets start by assumign that these are both set to whatever terrain_or_low_angle was
- For masking the sea bit away, use Fractional Cover Percentiles as a base:
                    "flags": {
                        "band": "land",
                        "product": "geodata_coast_100k",
                        "ignore_time": True,
                        "ignore_info_flags": [],
                    },

pindge:
Merged in through refactor

PR #645, opened 2021-01-21
split mega ows_cfg file to manageable size

pindge:
spitting ows_cfg into files less than 1k line


SpacemanPaul (changes requested):
Could be split further, but definitely an improvement.

SpacemanPaul (approved):

PR #651, opened 2021-01-25
finalise ows cfg refactoring

pindge:
-  fix errors
- apply black formating
- refine more layers to make files smaller and modules more reusable
- add more layers
- remove unused sections
- re-organise and re-grouping layers
- reduce any files to less than 500 lines


SpacemanPaul (approved):
Wow, this is a massive PR.  But it's a massive improvement too.

PR #652, opened 2021-01-25
only build when changes to ows

pindge:
- only build new docker image when there is change to ows.
@alexgleith @omad unless you guys depend on docker images for other stuff?


whatnick (approved):
💯

PR #654, opened 2021-01-28
add dev inventory.json

pindge:
- check ows-cfg against `inventory.json` to identify the missing layers or styles
- ows `datacube-ows-cfg-parser` doesn't return an error on mismatch for inventory so the GHA CI is not returning a fail
- another PR will be raisedafter `datacube-ows-cfg-parser` is coded to return an error if inventory isn't matching.


SpacemanPaul (commented):
Makes sense to me.

whatnick (approved):

PR #657, opened 2021-01-29
bex adding tas data to terria wit

BexDunn:
updated the terria json config to include the three tasmanian datasets for demo to tas stakeholders. Tested in a clean terria map instance, everything seems to be working :)


pindge:
Look how beautiful this is

pindge (approved):

PR #660, opened 2021-02-08
remove log_mad*

pindge:
```
The config issue is that the three style_tmad_*dev  styles should be deleted (keeping the style_tmad_*dev_std  styles).
```

### DIFF
--- a/dev/services/wms/ows_refactored/tmad/ows_tmad_cfg.py
+++ b/dev/services/wms/ows_refactored/tmad/ows_tmad_cfg.py
@@ -7,11 +7,11 @@ bands_tmad = {
 }
 
 style_tmad_sdev_std = {
-    "name": "arcsec_sdev",
+    "name": "log_sdev",
     "title": "SMAD",
     "abstract": "Good for cropland and forest",
     "index_function": {
-        "function": "datacube_ows.band_utils.single_band_arcsec",
+        "function": "datacube_ows.band_utils.single_band_offset_log",
         "mapped_bands": True,
         "kwargs": {"band": "sdev", "scale_from": [0.017, 0.15], "scale_to": [0.0, 4.0]},
     },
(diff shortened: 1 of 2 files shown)


dev/services/wms/ows_refactored/tmad/ows_tmad_cfg.py
```diff
@@ -5,109 +5,13 @@
...
-}
 
 style_tmad_sdev_std = {
-    "name": "arcsec_sdev",
+    "name": "log_sdev",
     "title": "SMAD",
     "abstract": "Good for cropland and forest",
     "index_function": {
-        "function": "datacube_ows.band_utils.single_band_arcsec",
+        "function": "datacube_ows.band_utils.single_band_offset_log",
```
SpacemanPaul, line 14: No, we want to keep arcsec as the function - which I guess means we have to change the name too.

SpacemanPaul (changes requested):
We want to keep the arcsec function for sdev, sorry.

dev/services/wms/ows_refactored/tmad/ows_tmad_cfg.py
pindge: so 3 `arcsec` styles?

pindge: because they are missing

SpacemanPaul: No bcdev and edev use log  - sdev uses arcsec.

I know it looks inconsistent, but it's how I tuned it with Dale.

SpacemanPaul: If you want it to look consistent, use sdev_scaled, edev_scaled, etc.

SpacemanPaul (approved):
Looks right to me.

PR #663, opened 2021-02-15
Remove "munged" from name of HAP

GypsyBojangles:
Remove "munged" from name of HAP


whatnick (approved):

pindge:
Thanks for doing both files,. but soon the mega file is going to be deleted.

pindge (approved):

PR #666, opened 2021-02-16
setup split config for prod

pindge:
- setup config for prod cfg mirroring dev
- comment out layers not supported in prod
- update inventory.json to reflect tmad fixes


SpacemanPaul (approved):
This is going to make config management a lot easier going forwards.

PR #667, opened 2021-02-17
add underscore to layername

pindge:
- add an underscore to layer names for consistency


GypsyBojangles (approved):

GypsyBojangles (changes requested):
You need to change the linked WCS coverage name too. Has dea.json been updated already?

pindge:
> You need to change the linked WCS coverage name too. Has dea.json been updated already?

wrong file

pindge:
> You need to change the linked WCS coverage name too. Has dea.json been updated already?

ok, fixed, change this round is to dev only, so only to `dea.json`

GypsyBojangles (approved):

PR #672, opened 2021-02-24
updated wofl colour scheme

erin-telfer:


dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
```diff
@@ -127,24 +127,69 @@
...
                         "noncontiguous": True,
                     }
                 },
-                "color": "#707070",
+                "color": "#ffffff",
+            },
+            {
+                "title": "Cloudy Terrain",
+                "abstract": "Sloped terrain and cloudy",
+                "flags": {"high_slope": True, "cloud": True},
```
SpacemanPaul, line 142: You will need to explicitly specify whether these flags are to be ANDed or ORed.

dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
```diff
@@ -127,24 +127,69 @@
...
+                "abstract": "Sloped terrain",
+                "flags": {
+                    "high_slope": True,
+                },
+                "color": "#776857",
+            },
+            {
+                "title": "Cloudy Water",
+                "abstract": "Wet and cloudy",
+                "flags": {"wet": True, "cloud": True},
```
SpacemanPaul, line 156: This one too - you must specify whether to use "AND" or "OR" logic when you include multiple flags.

SpacemanPaul (changes requested):
Please explicitly specify whether to use "AND" or "OR" logic when combining more than one flag predicate.

Also, we should consider the changes that will happen with Collection 3 - where the sea bit has to be pulled in from another product, meaning that we cannot colour-code it - only mask it.  Therefore the C3 wofs data has ALL sea pixels transparent - perhaps we should make this C2 data match, seeing as we will be changing soon anyway?

dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
```diff
@@ -127,24 +127,69 @@
...
+                "flags": {
+                    "cloud": True,
+                },
+                "color": "#d8d7d6",
             },
             {
                 # Possible Sea Glint, also mark as invalid
                 "title": "",
                 "abstract": "",
                 "flags": {"dry": True, "sea": True},
```
MatthewJA, line 185: This one too surely @SpacemanPaul?

SpacemanPaul: Yes, I gave up after two and left a general comment.  I'm honestly not sure how this is interpreted, which is a good reason to fix it now.  (I would have to trace through the code - the documentation says it should be an error.)

MatthewJA: Just noting this one in particular because it was in the codebase prior to this PR!

dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
```diff
@@ -127,40 +127,97 @@
...
+                "color": "#4f81bd",
             },
             {
                 "title": "Dry",
                 "abstract": "Dry",
                 "flags": {
-                    "dry": True,
-                    "sea": False,
+                    "and": {
+                    "dry": True, "sea": False
```
SpacemanPaul, line 215: We've already matched (dry and sea) so this could just be (dry).

But this is probably clearer to read.

SpacemanPaul (approved):
This is fine.  But note that the legend will change when we switch to C3 (as all sea:True pixels will be transparent.)

PR #677, opened 2021-03-01
fix c3 landsat styles

pindge:
- fix panchromic style title
- fix c3 false color range


SpacemanPaul (approved):
Perfect.   Well spotted.  ;)

PR #684, opened 2021-03-02
remove borken config for wet only

pindge:
- revert wofs wet only style, 
- align c2 style wet only

### DIFF
--- a/dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
+++ b/dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
@@ -233,7 +233,7 @@ style_wofs_obs_wet_only = {
                     }
                 },
                 "color": "#707070",
-                "mask": True,
+                "alpha": 0.0,
             },
             {
                 # Possible Sea Glint, also mark as invalid
@@ -241,7 +241,7 @@ style_wofs_obs_wet_only = {
                 "abstract": "",
                 "flags": {"dry": True, "sea": True},
                 "color": "#707070",
-                "mask": True,
+                "alpha": 0.0,
             },
             {
                 "title": "Dry",
@@ -251,7 +251,7 @@ style_wofs_obs_wet_only = {
                     "sea": False,
                 },
                 "color": "#D99694",
-                "mask": True,
+                "alpha": 0.0,
             },
             {
                 "title": "Wet",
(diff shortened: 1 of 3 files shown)


pindge:
@GypsyBojangles 

This was the style before, apart from `sea` and `water` every other bit was marked as `mask: true`, on terria they were not showing, however them on the config caused ows to generate that bit for the legend, so I just removed the `mask: true` config

@SpacemanPaul , so i guess its either ows shouldn't generate legend if `mask: true` in `valuemap` or this `wet only` style is not configured correctly.
```
style_wofs_obs_wet_only = {
    "name": "wet",
    "title": "Wet Only",
    "abstract": "Wet Only",
    "value_map": {
        "water": [
            {
                "title": "Invalid",
                "abstract": "Slope or Cloud",
                "flags": {
                    "or": {
                        "terrain_or_low_angle": True,
                        "cloud_shadow": True,
                        "cloud": True,
                        "high_slope": True,
                        "noncontiguous": True,
                    }
                },
                "color": "#707070",
                "mask": True,
            },
            {
                # Possible Sea Glint, also mark as invalid
                "title": "",
                "abstract": "",
                "flags": {"dry": True, "sea": True},
                "color": "#707070",
                "mask": True,
            },
            {
                "title": "Dry",
                "abstract": "Dry",
                "flags": {
                    "dry": True,
                    "sea": False,
                },
                "color": "#D99694",
                "mask": True,
            },
            {
                "title": "Wet",
                "abstract": "Wet or Sea",
                "flags": {"or": {"wet": True, "sea": True}},
                "color": "#4F81BD",
            },
        ]
    },
}
```

SpacemanPaul:
I believe it should not generate legend if mask is True.  If it doesn't (which would not surprise me), open it as an ows issue.

dev/services/wms/ows_refactored/wofs/ows_wofs_cfg.py
```diff
@@ -130,23 +130,23 @@
                 "title": "",
                 "abstract": "",
                 "flags": {"noncontiguous": True,},
-                "mask": True,    
+                "mask": True,
```
SpacemanPaul, line 133: May as well fix it here too.

SpacemanPaul (changes requested):

SpacemanPaul (approved):
Looks good.

Issue #681, opened 2021-03-02
Collection 3 ARD Style Questions


GypsyBojangles:
1. Should we ONLY display contiguous data for all C3 ARD styles? nbart_contiguity = true
2. Should indices be cloud masked? FMASK in ['valid', 'snow', 'water']
3. Should we include additional cloud masked styles for True and False colour? (and maybe more)
4. Should indice style names be of the form: 'NDVI - Red, NIR' (current OWS behaviour) or 'Normalised Difference Vegetation Index' (current GSKY behaviour)

For reference:
https://explorer.dev.dea.ga.gov.au/products/ga_ls8c_ard_3
https://explorer.dev.dea.ga.gov.au/products/ga_ls7e_ard_3
https://explorer.dev.dea.ga.gov.au/products/ga_ls5t_ard_3

robbibt:
> Should we ONLY display contiguous data for all C3 ARD styles? nbart_contiguity = true

I think this makes sense, and should account for nodata as well as help clean up some of the ugly results you get near the edge of Landsat paths and rows. There are however issues with the current 20 m resolution Sentinel-2 contiguity bands not being fit-for-purpose for masking 10 m Sentinel-2 bands (I'm going to email this out more widely later this week), so there's a chance a different approach might be required for Sentinel-2 down the track. 

> Should indices be cloud masked? FMASK in ['valid', 'snow', 'water']

Yes. Cloud masking these indices would be a great improvement over the current approach with can lead to very misleading visualisations of the indices under cloud (e.g. the MNDWI water index which maps clouds as a strong water signal).

> Should we include additional cloud masked styles for True and False colour? (and maybe more)

I think this would overcomplicate things - ideally we want to keep the number of options to a minimum where possible. Clouds in the True and False colour layers are easily recognizable visually, so I don't think need to be cloud-masked unlike the indices where clouds can be very difficult to distinguish from non-cloud. 

> Should index style names be of the form: 'NDVI - Red, NIR' (current OWS behaviour) or 'Normalised Difference Vegetation Index' (current GSKY behaviour)

I lean on the side of the GSKY behaviour here because it gives the user a lot more context; e.g. they might not know what "NDVI" is, but they will know they want a view of the data that focuses on "vegetation". That said, there are advantages in having the spectral bands info included too for indices like MNDWI/NDWI that vary by a single band. Would `Normalised Difference Vegetation Index (Red, Nir)` be too verbose? 

[...]

PR #686, opened 2021-03-02
Change "Simple RGB" to "True Colour"

GypsyBojangles:
Change "Simple RGB" style to "True Colour" as per advice in #681


pindge:
need to update `inventory.json` for it to pass test

pindge (approved):

PR #705, opened 2021-03-17
set fmask for c3 ls

pindge:
#681


dev/services/wms/ows_refactored/c3/ows_c3_cfg.py
```diff
@@ -75,6 +77,14 @@
...
     },
     "scale_range": [0.0, 3000.0],
+    "pq_masks": [
+        {
+            "band": "oa_fmask",
+            "enum": 0,
+            "invert": True,
+        }
+    ],
+    "additional_bands": ["oa_fmask"],
```
SpacemanPaul, line 87: You shouldn't need the "additional_bands".  It can figure out it needs the band from the data it has.

I only suggested it as a workaround while I was debugging.

dev/services/wms/ows_refactored/c3/ows_c3_cfg.py
```diff
@@ -87,6 +97,14 @@
...
```
SpacemanPaul, line 107: Remove additional bands.

SpacemanPaul (changes requested):

SpacemanPaul (approved):

PR #709, opened 2021-03-18
update c3 fmask values

pindge:


SpacemanPaul (approved):
I think this will do it.  But I'm keeping my fingers crossed because I've thought that before...

PR #713, opened 2021-03-19
update legend and WO text

erin-telfer:
updated C3 WO file to include legend size details and updated the abstract text


pindge (approved):
Looks great/!

PR #726, opened 2021-04-13
Add updated DEA Coastlines config as backup option for release

robbibt:
This PR adds the updated DEA Coastlines config to the current version 7 of Terria on DEA Maps. This is a backup in case version 8 of Terria is not stood up prior to tomorrow's release.

@GypsyBojangles @pindge If you get a chance, it would be great if this PR could be reviewed as a priority so we have time to verify it works correctly. Thankyou!

### DIFF
--- a/dev/terria/dea-maps.json
+++ b/dev/terria/dea-maps.json
@@ -1013,61 +1013,60 @@
          "preserveOrder": true,
          "items": [
             {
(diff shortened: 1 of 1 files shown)


pindge:
Are you using an IDE, can u apply prettify?

pindge (approved):
I will deal with the formating as a separate pr

PR #731, opened 2021-04-16
C3 ARD abstract as per CMI

GypsyBojangles:
As per discussion with DEA, update ARD layer abstracts to reflect CMI "What this product offers" sections as well as link back to CMI.


pindge (approved):
LGTM,

PR #733, opened 2021-04-16
C3 WO Abstract as per CMI

GypsyBojangles:
As per recent discussion with DEA teams about metadata conventions, Layer abstract is now copied from "What this product offers" section of CMI and includes hyperlink to CMI record.


pindge (approved):
LGTM

PR #740, opened 2021-05-09
collections.csv

ramagopr:
Add collection list as a CSV to Github. This file will be used for the data ingestor of dea-access


mmochan (approved):
LGTM!

PR #748, opened 2021-05-14
Added S2 WO Alchemist config

MatthewJA:
This PR adds an Alchemist config file for Sentinel 2 definitive WO.

### DIFF
--- a/dev/services/alchemist/ga_s2_wo_3/s2_wo.alchemist.yaml
+++ b/dev/services/alchemist/ga_s2_wo_3/s2_wo.alchemist.yaml
@@ -55,5 +55,5 @@ output:
 
 processing:
   dask_chunks:
-    x: 600
-    y: 600
+    x: 6144
+    y: -1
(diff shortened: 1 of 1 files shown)


dev/services/alchemist/ga_s2_wo_3/s2_wo.alchemist.yaml
```diff
@@ -0,0 +1,59 @@
...
+    collection_number: 3
+    naming_conventions: dea_s2_derivative
+
+  properties:
+    dea:dataset_maturity: interim
+
+processing:
+  dask_chunks:
+    x: 600
+    y: 600
```
alexgleith, line 59: We should change this.

```suggestion
    x: 6144
    y: -1
```

Since the shape is `"shape": [10980, 10980]`, and blocksize is 1024, let's do it in halves, like we do the Landsat data. See: https://github.com/opendatacube/datacube-alchemist/blob/main/examples/c3_config_wo.yaml

alexgleith (approved):

PR #751, opened 2021-05-17
Update the collections.csv 

ramagopr:
tUpdate the collections.csv to enable only 3 collections that are related to Collection 3


mmochan (approved):
LGTM!

PR #759, opened 2021-06-07
add yaml linting

pindge:
- add yaml linting

### DIFF
--- a/.github/workflows/yaml_linting.yaml
+++ b/.github/workflows/yaml_linting.yaml
@@ -1,0 +1,27 @@
+name: Lint YAML files
+
+on:
+  push:
+    paths:
+      - "products/**"
+      - .github/workflows/yaml_linting.yaml
+  pull_request:
+    branches: [master]
+    paths:
+      - "products/**"
+      - .github/workflows/yaml_linting.yaml
+
+jobs:
+  yamllint:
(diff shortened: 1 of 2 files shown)


.github/workflows/yaml_linting.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+      - name: Check file naming
+        run: |
+          files=$(find products/**/*.yml | wc -l)
+          if [ $files -gt 0 ]; then echo "File found with .yml extension, please rename with .yaml" && exit 1; fi
+
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v3
+        with:
+          file_or_dir: dev
```
omad, line 28: Why only on `dev` and not on `products`? Also, `dev` isn't included in the GitHub actions paths configuration.

omad (changes requested):
Hi Pin, more linting sounds good to me, just curious about the decisions on what to check in the different directories.

Can you also please clean up the commits before we merge, it only needs one or two.

pindge (commented):
> Hi Pin, more linting sounds good to me, just curious about the decisions on what to check in the different directories.
> 
> Can you also please clean up the commits before we merge, it only needs one or two.

yes, will

PR #761, opened 2021-06-07
add yamllint

pindge:
- add yamllint action
- remove tab


.github/workflows/yaml_linting.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: 'Yamllint'
+        uses: karancode/yamllint-github-action@master
+        with:
+          yamllint_file_or_dir: products/**/*.yaml
+          yamllint_strict: false
+          yamllint_config_filepath: yamllint_cfg.yml
+          yamllint_comment: true
+        env:
+          GITHUB_ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
omad, line 33:
```suggestion

```
This shouldn't need any secrets or tokens.

omad (changes requested):

.github/workflows/yaml_linting.yaml
pindge: Its useful to highlight the bad bit in the yaml.

omad: Oh okay, this action can also post a comment on a PR, I didn't realise that.

omad (approved):

PR #768, opened 2021-06-17
Remove opacity from WPS

GypsyBojangles:
As per advice from Nick Forbes-Smith, WPS elements shouldn't have an opacity fields.


robbibt (approved):
Looks good to me

PR #803, opened 2021-07-30
C3 Web service production release

GypsyBojangles:
Sync Dev and Production WMS configs for release of C3 web services and features
- Add Preserve User date order to C3 ARD styling for upcoming release
- Add delta styles to all Sentinel-2 layers
- Add FMask styles to all Sentinel-2 layers
- Made zoom and cache behaviour consistent between NRT and ARD
- Updated Sentinel-2 Layers to have consistent indices names
- Update inventory


pindge (approved):
Ah, easy, only an extension to an existing product layer

PR #804, opened 2021-07-30
Added progressive cache control to all products.

SpacemanPaul:
Progressive cache control is awesome.  We should be progressive cache controlling all the things.


pindge:
This pr is killing https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/dev_only/ows_hap_cfg.py 
```
{'values_changed': {"root['total_layers_count']": {'new_value': 83, 'old_value': 84}}, 'iterable_item_removed': {"root['layers'][44]": {'product': ['historical_airborne_photography'], 'styles_count': 1, 'styles_list': ['simple_gray']}}}
```

pindge:
This is the error:

```
 [2021-08-02 00:59:10,849] datacube_ows.ows_configuration [None] [ERROR] Could not parse layer: Dataset cache rule min_datasets value exceeds the max_datasets limit in layer historical_airborne_photography.
```

dev/services/wms/ows_refactored/ows_reslim_cfg.py
```diff
@@ -1,42 +1,47 @@
 # Reusable Chunks 1. Resource limit configurations
 
+dataset_cache_rules = [
+    {
+        "min_datasets": 5,
+        "max_age": 60 * 60 * 24,
+    },
+    {
+        "min_datasets": 9,
```
pindge, line 9: this must be a typo, it should be `max_datasets`

pindge (changes requested):

dev/services/wms/ows_refactored/ows_reslim_cfg.py
SpacemanPaul: Actually no, in the cache rules it is "min_datasets".

SpacemanPaul: You're not far off the problem though.

pindge (approved):

PR #808, opened 2021-08-04
The odc.stats WOfS summary running config file

supermarkion:
This yaml file the CONFIG file with https://github.com/opendatacube/odc-tools/pull/392


omad (approved):
I think we need to do some restructuring of this repository, but, for now, we are indeed running odc-stats inside the _development_ account, so this is fine.

Thank you so much Sai!

PR #809, opened 2021-08-10
move native_crs and native_resolution from wcs section to layer

pindge:
- update config to align with latest ows
- add `native_crs` and `naive_resolution` to `insar` products.


SpacemanPaul (approved):
Yep, that should do it.

PR #810, opened 2021-08-11
fix column NCI file location for wo/fc collections

ramagopr:
modify nci file location for derivatives collection fc and wo


omad:
Thanks for your PR!

However, I've checked on the NCI, and the correct base directory is still `/g/data/jw04/ga`.

Please reopen if I've misunderstood something.

PR #815, opened 2021-08-24
sentinel-2 NRT provisional ows configs

uchchwhash:
- added ows layers
- added them to the root config
- added to inventory


pindge (approved):
Fixed this change, see my commits. All passing now,

PR #822, opened 2021-08-26
deploy provisional ard to prod

pindge:
- deploy change in dev to prod
- deploy provisional_nrt


uchchwhash (approved):
I'm not too worried about the wording right now. We can change it later.

PR #828, opened 2021-09-05
Add initial draft of terria-cube-v8 catalog

GypsyBojangles:
Add initial draft of terria-cube-v8 catalog


robbibt (approved):
This looks fantastic to me - Landsat NRT, exciting!

These names are getting a bit clunky, but I don't have any better suggestion. I guess the idea is that the `Provisional` layers are a temporary thing that aren't intended for public eyes, so it's probably not a problem.

> DEA Surface Reflectance (Landsat) Provisional - near real time

Issue #834, opened 2021-09-10
terria catalog isn't applying dateformat correctly


pindge:
![image](https://user-images.githubusercontent.com/24644475/132812398-a8c20827-e4a1-4358-aa6d-126646ba090c.png)

Land cover is an annual product when the catalog `dateformat` is configured correctly it only shows the `YEAR`, but right now its showing `dd/mm/yyyy time` 

```
"dateFormat": {
      "currentTime": "'Year: 'yyyy",
      "timelineTic": "yyyy"
}
```

GypsyBojangles:
So that dateFormat syntax doesn't work in v8, we will need to get advice from TerriaJS team on what that should be,

nf-s:
Hi @GypsyBojangles 

The `dateFormat` just a string property now, `timelineTic` isn't supported.

So the following should work:

```json
"dateFormat": "'Year: 'yyyy"
```

GypsyBojangles:
Thankyou @nf-s !

GypsyBojangles:
Added correct syntax to terria-cube catalog in PR #835

PR #835, opened 2021-09-10
Add annual date filter to annual products

GypsyBojangles:
Add annual date filter to annual products
fix issue #834


GypsyBojangles:
So that dateFormat syntax doesn't work in v8, we will need to get advice from TerriaJS team on what that should be, Do not merge.

pindge (approved):

robbibt (approved):
Looks nice to me! 
image

PR #853, opened 2021-09-23
move the top-level provisional ARD products to sensor-specific folders

uchchwhash:


robbibt (approved):
This looks perfect to me, thanks Imam!

PR #859, opened 2021-09-28
move feature info script to correct layer

bellemae:


SpacemanPaul (approved):
I think this should work.

PR #863, opened 2021-09-29
Add Annual FCP Product Definition

GypsyBojangles:
Add Annual FCP Product Definition according to specification.


products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
```diff
@@ -0,0 +1,79 @@
...
+description: Geoscience Australia Landsat Fractional Cover Percentile Calendar Year Collection 3
+license: CC-BY-4.0
+metadata_type: eo3
+
+metadata: 
+  product:
+    name: ga_ls_fc_pc_cyear_3
+  properties:
+    odc:file_format: GeoTIFF
+    odc:product_family: fc
```
alexgleith, line 11: I don't usually include these fields and just rely on matching by name.

products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
```diff
@@ -0,0 +1,79 @@
...
+        4
+        5
+        6
+        7
+      ]
+      values:
+        0: insufficient observations wet
+        1: insufficient observations dry
+        2: sufficient observations
+      description: Quality Assurance band for Mangrove Canopy Cover
```
alexgleith, line 79: Mangroves?

alexgleith (changes requested):
I think you might want to store this here: https://github.com/GeoscienceAustralia/digitalearthau/tree/develop/digitalearthau/config/eo3/products-aws

But I have two changes that I have recommended.

emmaai (approved):

products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
GypsyBojangles: I think the product will be named Mangrove Canopy Cover, not Mangroves Canopy Cover - @emmaai should this be a plural?

products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
GypsyBojangles: It seems we did include this in our WO C3 product definition. @omad do you want me to remove them?

products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
alexgleith: Sorry, I wasn't fussing about the plural. More that is said mangrove! If that's what the band does, then fine.

robbibt: According to the naming convention sheet, this is now "DEA Mangroves" (at least in short name format!)

GypsyBojangles: I feel like the plural will disappear when we update the specification document soon. I am going to resolve this conversation for now.

products/fractional-cover/ga_ls_fc_pc_cyear_3.yaml
omad: Most of the products do include some extra information, but I don't think it matters one way or another.

I'd be inclined to remove the `odc:file_format` at least, in case we start saying `COG` or switching to `zarr` or something crazier. But, neither of them are sufficient to differentiate which product a dataset will be, so.... it's safe to remove them.

GypsyBojangles: Removed

alexgleith (approved):

PR #865, opened 2021-09-29
replace 0s with n/a in label mapping

bellemae:
remove superfluous 0s in value maps

### DIFF
--- a/dev/services/wms/ows_refactored/land_and_vegetation/landcover/lc_class_desc.py
+++ b/dev/services/wms/ows_refactored/land_and_vegetation/landcover/lc_class_desc.py
@@ -1,5 +1,5 @@
 level3_labels = {
-    0: "No data",
+    'n/a': "No data",
     111: "Cultivated Terrestrial Vegetation",
     112: "Natural Terrestrial Vegetation",
     123: "Cultivated Aquatic Vegetation",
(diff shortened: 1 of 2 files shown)


SpacemanPaul (changes requested):
I don't think this is going to work, sorry.

dev/services/wms/ows_refactored/land_and_vegetation/landcover/lc_class_desc.py
```diff
@@ -1,5 +1,5 @@
 level3_labels = {
-    0: "No data",
+    'n/a': "No data",
```
SpacemanPaul, line 2: This won't work.  We're better off leaving this file as it was, but .....

dev/services/wms/ows_refactored/land_and_vegetation/landcover/ows_c2_lc_cyear_cfg.py
```diff
@@ -19,7 +19,6 @@
     "abstract": "Standardised colouring of Level 3 land cover classes",
     "value_map": {
         "level3": [
-            {'title': "", 'abstract': "", 'values': [0], 'color': '#FFFFFF', 'alpha': 0, },
```
SpacemanPaul, line 22: Leave these in too.

PR #870, opened 2021-09-30
add tmads to terria-cube and deamaps

GypsyBojangles:
Add TMADs to terria-cube and deamps.
This kind of got lost a year or two ago, but a bunch of work went into it and the web services exist.


robbibt (approved):
Looks great!

PR #915, opened 2021-10-14
Remove the time dimension from the LS C3 product definitions.

tebadi:


tebadi:
> Do you know if this affects or breaks anything? Curious more than anything, it definitely doesn't belong there.
> 
> Thanks!

No, but I'm going to test indexing on it and see if it breaks anything.

PR #924, opened 2021-10-18
change waterbodies charts to "lineandpoint" as per Nanda's advice

GypsyBojangles:
As per Nanda's advice on Slack:
"
Hey guys the change is to enable dots for waterbodies chart as was requested in this issue - https://github.com/TerriaJS/de-australia-map/issues/72
the catalog change is to add an attribute to the <chart> component in the featureinfo template for waterbodies item. The new attribute is chart-type='lineAndPoint'."


robbibt (approved):
Looks good!

PR #928, opened 2021-10-19
align text and folder structure for ows prod

pindge:
align text changes


pindge:
> Hey @pindge - this looks like all the changes from dev have been moved over correctly. Are the test fails OK?

i am fixing the errors, thanks!

PR #931, opened 2021-10-20
Waterbodies fix for terria

vnewey:
Change DEA Waterbodies on Terria to point to the DEA geoserver not hotspots. Change for both dev and prod on Terria


robbibt (changes requested):
Hey Ness, looks great - we probably want to update the DEA prod geoserver here too:

https://github.com/GeoscienceAustralia/dea-config/blob/waterbodies_dev_fix/dev/terria/terria-cube-v8.json#L3815

robbibt (approved):
Looks great!

PR #934, opened 2021-10-21
Change Waterbodies to point to dea geoserver wms

vnewey:
Change Waterbodies on dea maps to point to dea geoserver wms


robbibt (approved):
Looks great!

PR #949, opened 2021-11-03
change s2 product back to c2 for wo config

erin-telfer:
Change s2 nrt product back to s2c2, as assumption about updating nrt s2 wo from c2 (s2*_nrt_granule) to c3 (ga_s2*m_ard_provisional_3) was incorrect. Not just as easy as changing product name.


MatthewJA (approved):
LGTM

PR #956, opened 2021-11-10
Mark v7 Terria configs as deprecated

robbibt:
This PR renames the old "dea-maps.json" and "dea.json" files to "dea-maps-v7_deprecated.json" and "terria-cube-v7_deprecated.json" for consistency with the new v8 names, and to make it clear that these are no longer used to configure our web services on DEA Maps or Terria Cube.


robbibt:
@pindge The linter is failing because the files have been renamed - what if we removed the specific references to those files from the list in the Github action? They aren't used anymore anyway, so I don't think they need to be specifically included

pindge (approved):

PR #962, opened 2021-11-18
add explorer product list that will be used for aws s3 storage stats

ramagopr:
Adding a list of explorer products for which S3 storage stats have to be collected.


pindge:
can u add usage for this file to the readme https://github.com/GeoscienceAustralia/dea-config/blob/master/workspaces/README.md

PR #965, opened 2021-11-23
Add sensible WCS resource limits.

SpacemanPaul:
WCS requests were previously unrestricted, which means that a large WCS request would time out and eventually crash the pod instead of failing immediately with a useful error message.

The following restrictions now apply to all WCS requests:

1) Limit on the size of the uncompressed data to 800Mb.
2) Limit on the total number of dataset file accessed to 30.


robbibt:
Hey @SpacemanPaul, I'm not sure how to test this, but those limits sound good to me generally. I assume this now produces an error if a user goes over the limit that could be caught inside TerriaJS as part of the clip and ship?

pindge (approved):

robbibt (approved):

SpacemanPaul:
@robbibt - that's the idea.

PR #969, opened 2021-11-25
Reduce WCS dataset limit slightly

SpacemanPaul:
Was 30, which resulted in ~40s queries.  Reducing to 24.


robbibt (approved):
Looks good!

PR #974, opened 2021-11-30
add ocean mask to c3 WO products

erin-telfer:
adding flag masking to c3 wo products to show land and mask ocean


SpacemanPaul (approved):
It's quite a large PR so I might have missed something, but looks ready to merge to me.

PR #980, opened 2021-12-02
Remove C2 GeoMAD and Publish C3 GeoMAD

GypsyBojangles:
Remove C2 Geomedian and TMAD from Terria catalogs
Add C3 GeoMad to DEA Maps


robbibt (approved):
Looks good to me - have tested all the layers. Looks like there was a decision to use "DEA GeoMAD" rather than the more generic "DEA Surface Reflectance Calendar Year"? (I remember there being a meeting where we settled on the second option, but that was a while ago so things probably changed since then).

One thing to note for later: it looks like the current WMS abstracts for these products (and all C3 products?) aren't separating the layer name title from the body text. It would be nicer if this could be formatted with a line break so they didn't run together into a single paragraph - a job for a different PR I think though

image

Issue #981, opened 2021-12-02
OWS WMS descriptions should include space between product name and "what this product offers"


robbibt:
It looks like the current OWS WMS descriptions for all our C3 Landsat-derived products (and our C1 Sentinel-2 products) aren't including a new line separator between the <Full Technical Product Name> and the <What this product offers> text. It would be nicer if this could be formatted with a line break so they didn't run together into a single paragraph. For example:

![image](https://user-images.githubusercontent.com/17680388/144351733-8c18c30a-83ca-4dfa-96ed-9038354eca7c.png)

The layers that this seems to affect include:

- DEA Surface Reflectance (Landsat)
- DEA Surface Reflectance (Sentinel-2 Near Real-Time)
- DEA Surface Reflectance (Sentinel-2)
- DEA GeoMAD (Landsat 8 OLI-TIRS)
- DEA GeoMAD (Landsat 7 ETM+)
- DEA GeoMAD (Landsat 5 TM)
- DEA Surface Reflectance (Landsat 8 OLI-TIRS)
- DEA Surface Reflectance (Landst 7 ETM+)
- DEA Surface Reflectance (Landsat 5 TM)
- DEA Surface Reflectance (Sentinel-2A MSI Near Real-Time)
- DEA Surface Reflectance (Sentinel-2B MSI Near Real-Time)
- DEA Surface Reflectance (Sentinel-2A MSI)
- DEA Surface Reflectance (Sentinel-2B MSI)
- DEA Fractional Cover (Landsat)
- DEA Water Observations (Landsat)

The following products correctly include a space below the product title:

- DEA Surface Reflectance (Landsat, Provisional)
- DEA Surface Reflectance (Landsat 8 OLI-TIRS, Provisional)
- DEA Surface Reflectance (Landst 7 ETM+, Provisional)
- DEA Surface Reflectance (Sentinel-2, Provisional)
- DEA Surface Reflectance (Sentinel-2A MSI, Provisional)
- DEA Surface Reflectance (Sentinel-2B MSI, Provisional)

![image](https://user-images.githubusercontent.com/17680388/144351980-6b5edbdb-357c-42a5-b317-1e70e81c1ad3.png)

robbibt:
Seems to be just a case of adding in an extra line break here:
https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/baseline_satellite_data/landsat/ows_c3_cfg.py#L89-L90

[...]

PR #984, opened 2021-12-03
OWS Abstract formatting

GypsyBojangles:
Apply consistent formatting to abstracts.

Attempt to fix issue: #981


robbibt (approved):
Looks great to me - fixed a couple of minor issues.

Will do another check on the layers themselves once this is merged

PR #985, opened 2021-12-06
deploy wofs in bulk to prod

pindge:
- deploy 4x wofs layers
- apply the structural change to wofs folder
- update metadata
- update lccs legend URL to production


pindge:
will create a clean pr

PR #990, opened 2021-12-09
Make Landcover Extents Prettier in Explorer

omad:
Adding the expected gridding into the Landcover Product definition should make it look prettier in Explorer.

Please confirm that Landcover uses the same tiling configuration as the Collection 2 Albers products.

image

image


jeremyh (approved):
I don't know if the numbers are what was used to generate the product, but I'm in favour of adding them.

PR #1012, opened 2022-01-23
enable mstp legend and fix intertidal extents legend display

pindge:
- make `mstp` legend display
- give `confidence_layer` some white space padding, so it doesn't look like its been cut off.

Current
legend

After change
download
 
- fix `confidence_layer` unit in dev


SpacemanPaul (approved):
Yep, this will address the issues.

PR #1023, opened 2022-01-31
apply refactored wofs

pindge:
## Change
- deploy 4x wofs c3 layers
- refactor wofs layer folder as per changes applied to dev
- update lccs dev legend URL to match prod
- move wofs c3 in inventory.json together


erin-telfer (approved):
While I haven't been through and checked off every update, overall it looks good and the detailed changes that I have examined look great. Thanks Pin!

PR #1024, opened 2022-01-31
Fix Collection 3 Fractional Cover

omad:
We whoopsed in December by merging a configuration change to Fractional
Cover which performs corrections on the output values of Landsat
8 FC. The whoops is because the code hasn't yet been released yet, and
so LS8 FC has been failing since then!


emmaai (approved):
My fault merging the config without merging the code.

omad:
> My fault merging the config without merging the code.

And my fault for both letting it happen and then doubly so for not noticing earlier. 😆

PR #1027, opened 2022-02-07
fc/tcw percentile config change

emmaai:
Move common configs into config file rather than carrying by the command line options every time.


omad (approved):
Nice!

Thanks Emma

PR #1032, opened 2022-02-09
readding WO to DEA Maps catalogue

erin-telfer:
fixed bracket/alignment errors that made it through linting tests.


erin-telfer:
I give up

PR #1035, opened 2022-02-14
Fix zoom_factor/zoom_level typo in reslims.

SpacemanPaul:
The resource_limit in prod for Sentinel-2 products was using "min_zoom_factor" instead of "min_zoom_level" resulting in ridiculously expensive queries being attempted, with timeouts and crashes resulting.

This should fix.


pindge (changes requested):
@SpacemanPaul , line 64 require attention

SpacemanPaul:
> @SpacemanPaul , line 64 require attention

No it doesn't.  Firstly it's not used in prod at all - it's only used for the HAP prototype in dev, where the current value is reasonable.

pindge (approved):
@SpacemanPaul , line 64 require attention

PR #1036, opened 2022-02-18
Update catalogue to reflect Waterbodies v1 and v2

vnewey:
Updating the Terria catalogue to reflect the v1 and v2


robbibt (approved):
This looks great @vnewey! Both layers work well for me

PR #1038, opened 2022-02-20
Update Waterbodies on dea maps

vnewey:
Point DEA Maps to the non-prod waterbodies webservices while the prod services are being updated

### DIFF
--- a/dev/terria/dea-maps-v8.json
+++ b/dev/terria/dea-maps-v8.json
@@ -1377,7 +1377,7 @@
                             "members": [
                                 {
                                     "type": "wms",
-                                    "name": "DEA Waterbodies v2",
+                                    "name": "DEA Waterbodies (version 2)",
                                     "url": "https://nonprod.geoserver.dea.ga.gov.au/geoserver/dea/wms",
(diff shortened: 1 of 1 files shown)


robbibt:
Hey @vnewey, a general question: do we need to include both the old and new versions on DEA Maps? I completely understand having them both in Terria Cube, but given DEA Maps is public facing perhaps we don't want to confuse users by giving them multiple options?

The other slight inconsistency with other DEA layers in DEA Maps is that we don't include version numbers (e.g. "v2") anywhere else in the catalogue. My personal preference would to just have the most recent version of Waterbodies in DEA Maps without the grain suffix, but if you did want to include both, this naming format might match the existing layers a little more:

- DEA Waterbodies (version 2)
- DEA Waterbodies (version 1, deprecated)

vnewey:
Hi @robbibt,
Thank you for your feedback! After talking to Claire and Bex, I have decided to keep version 1 in the DEA maps catalogue for the short term. I have changed the naming format to match your suggestion.
Thanks for your help!

BexDunn (approved):
Looks good Ness. Changes are good, and only comment was updating the spelling of 'depricated' as chatted about on slack in the WMS layer descriptions.

PR #1040, opened 2022-02-22
Waterbodies v2

vnewey:
Switching Waterbodies catalogue item to point to the newly updated wms on the dea prod geoserver.
Also, changing Waterbodies on terria and dea maps to use a png for the legend.


BexDunn:
image 
Looks good Ness. Do you want to update the legend to match v1 to distinguish between wet pixel count and wet pixel percentage more clearly?

BexDunn (approved):

PR #1041, opened 2022-02-23
update WIT in dev so that it matches prod and loads

BexDunn:
The Wetlands Insight Tool catalogue entry in dev is broken in Terria. This PR updates the catalogue entry by using the catalogue entry for prod, so that folk looking for WIT will find it if they look in the dev entry rather than the prod entry.


vnewey (approved):
Looks good!

PR #1049, opened 2022-04-06
LC C3 product definition

eloise-b:


tebadi (commented):
Looks good!

margaretharrison (commented):
Looks good, no changes from me

erin-telfer:
Hi @eloise-b, I can't be certain that all metadata properties are set for C3, but nothing is obviously missing. Keep an eye on when adding product to ODC and if the product fails, I suggest reaching out to core or Jeremy. 

To the best of my knowledge looks ok :)

erin-telfer (approved):

PR #1050, opened 2022-04-06
remove c2 product ows layers and catalog

pindge:
Removing selected c2 product layers from ows dev and prod, summary below:
- removing `28` layers 
- update terria and dea-map catalog


CEKrause (approved):
I can't comment on the detail, but the layers being deleted looks correct to me. Thanks @pindge

PR #1062, opened 2022-05-12
fix fcp ows style capitals and legend

emmaai:


pindge:
@emmaai can you also move https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/land_and_vegetation/ows_category_root_cfg.py#L13-L16 to https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/land_and_vegetation/fc/ows_fc_cfg.py#L11 so it sits in the folder

pindge (approved):

PR #1063, opened 2022-05-12
add fcp cyear to catelog

pindge:
- move `fc rgb` style to top of the list so its loaded as default layer
- add `fcp cyear` to terria catalog


robbibt (approved):
Looks good to me!

PR #1082, opened 2022-06-22
Add GeoMAD product definitions for financial years.

tebadi:


jmettes (approved):

tebadi:
> Looks good. One thing I'm not sure of is checking whether the scale factors are correct

@omad could you please check this?  These are the same in calendar years and financial years.

omad:
@jmettes @tebadi I don't understand. These don't contain any scale factors?

tebadi:
> @jmettes @tebadi I don't understand. These don't contain any scale factors?

I thought Jono was referring to resolution and tile size but he can clarify himself.

jmettes:
Yes, tile size and resolution, in conjunction to discussions about how geomedian/odc-stats was oversampling. I checked, and I think it's correct.

PR #1093, opened 2022-08-01
Update calendar year mangroves odc-stats config to 3.0.0

jmettes:
Collection 3 Mangroves was mistakenly deployed to production under version 2.0.2. Correcting OWS and odc-stats to have correct version of 3.0.


jmettes:
I also found some references to Mangroves 2.0.2 in:
* https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/terria/dea-maps-v7_deprecated.json
* https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/terria/terria-cube-v7_deprecated.json

Can these be deleted? They seem to be deprecated

pindge (approved):

PR #1094, opened 2022-08-09
update config for summary products

emmaai:
- collect all configs in one place
- add cloud/shadow dilation config
- fix geomedian bands name in thumbnail


tebadi (approved):
Looks good!

PR #1095, opened 2022-08-18
setup cmi checker

pindge:


omad:
Bye bye, ye old CMI. (to the tune of American Pie)

PR #1096, opened 2022-09-05
add transform code in stats config

emmaai:
To support the explicit transform path in reproject for the summary products, and it will be the same as the current default with gdal. Ref. https://github.com/opendatacube/odc-stats/pull/63


tebadi (approved):

tebadi:
Would be good to get Robbi's review before you merge.

robbibt (approved):
This looks great to me - a nice workaround for this issue too. 😃

PR #1102, opened 2022-10-26
Add s2cloudless_mask to ga_s2*m_ard_3 layer

jmettes:


robbibt:
Hey @jmettes, would it be possible for us to also add the "s2cloudless_prob" probability layer to OWS as well? I think it will be really valuable to be able to inspect the actual cloud probabilities as well as the integer mask in OWS, especially to help understand areas where we get weird results.

jmettes:
> Hey @jmettes, would it be possible for us to also add the "s2cloudless_prob" probability layer to OWS as well? I think it will be really valuable to be able to inspect the actual cloud probabilities as well as the integer mask in OWS, especially to help understand areas where we get weird results.

Yep, I can do that. I'm thinking of using a [ramp style](https://datacube-ows.readthedocs.io/en/latest/style_howto_color_ramp.html) of two values (value: 0, opacity: 0) to (value: 1, opacity: 1).

Might make a separate PR for this tho.

pindge (approved):

PR #1103, opened 2022-10-28
Fix s2cloudless layer mapping

jmettes:
Fix the s2cloudless layer to actually show s2cloudless


robbibt (approved):
That looks like it's the problem!

jmettes:
Thanks for reporting @robbibt!

PR #1104, opened 2022-10-28
Add Sentinel-2 C3 to dev terria catalogue

jmettes:
Unsure about this one - manually updated the json, which feels super gross. Untested too.

No idea about the embedded template HTML - it's probably wrong, just copied from other S2 NRT ones.

Also no idea about the `id`s - I just made them up.


pindge (approved):
👍

robbibt:
Hey @jmettes, this will update DEA Maps itself, not dev Terria Cube. I think you should be able to just cut the change here and move it to the "dev" section of "terria-cube-v8.json" in this same directory.

As for workflow, manually editing the config files and using a random ID is fine, that's how I've always done it in the past. 😃

robbibt:
There's some more details in the readme here: https://github.com/GeoscienceAustralia/dea-config/tree/master/dev/terria#readme

For unknown reasons, all the Terria configs (both "dea-maps-v8.json" and "terria-cube-v8.json" are located inside the directory "dev/terria/"; it'd make much more sense for "dea-maps-v8.json" to be located somewhere like "prod/terria/"

jmettes:
Moved changes to `terria-cube-v8.json`. Also updated some embedded HTML which were resulting in invalid JSON errors

jmettes:
Definitely agree on `dea-maps-v8.json` needing to be in `prod/terria` instead of `dev/terria` - that had me really confused.

Will need to stage that change alongside with DEA Maps pointing to it.

robbibt:
This looks great @jmettes! I made a tiny change to the layer title name format to match the Provisional layers.

One last thing that isn't working as intended: clicking on the older S2 C1 layers (e.g. "DEA Surface Reflectance (Sentinel-2)") brings up a Feature Info pop-up with values for each of the satellite bands, and a chart showing a spectral curve:

image

However doing the same thing for the new S2 C3 layers produces empty satellite band data and no chart:

image

robbibt (approved):
Fixed the chart/click issue (the FeatureInfoTemplate was pointing to NBAR layers which no longer exist on the Collection 3 data). Everything looks good to me!

PR #1105, opened 2022-11-01
Add s2cloudless_prob

jmettes:
Add basic `color_ramp` layer for `s2cloudless_prob`
image


robbibt (approved):
Looks great to me! Will have a closer look once it's live on dev OWS

PR #1111, opened 2022-11-08
Add pansharpened RGB style to Landsat 8 in dev OWS

robbibt:
This is an experimental test of pansharpening in OWS, using the demo code in this notebook: https://gist.github.com/robbibt/285bd20d25b9257dad3b162d965282f5

This approach uses Brovey pan-sharpening, a simple method that divides each RGB band by the sum of all band values, then multiplies the output by the pan band itself.

Very open to ideas/thoughts on improving this approach - consider this a very early first draft. 😃


dev/services/wms/ows_refactored/baseline_satellite_data/landsat/style_c3_cfg.py
```diff
@@ -550,6 +550,37 @@
...
+    "name": "true_colour_pan",
+    "title": "True Colour - Pansharpened",
+    "abstract": "True-colour image pansharpened to 15 metre resolution using Brovey pansharpening",
+    "additional_bands": ["nbart_red", "nbart_green", "nbart_blue", "nbart_panchromatic"],
+    "components": {
+        "red": {
+            "function": "ows_refactored.baseline_satellite_data.landsat.utils.brovey_pansharpen",
+            "kwargs": {
+                "output_band": "nbart_red",
+                "scale_from": [50, 600],
```
SpacemanPaul, line 563: Is there a reason we are only scaling from [50,600] instead of [0,3000] which is used for the standard true colour style?

SpacemanPaul (commented):
Questioning the scaling which is very different to the unsharpened true colour style.  Otherwise looks good!

dev/services/wms/ows_refactored/baseline_satellite_data/landsat/style_c3_cfg.py
robbibt: It's a quirk of the Brovey method, it doesn't return the pansharpened data with the original scaling. I've used histogram stretching in the past to revert it back to a "normal" value distribution, but didn't want to potentially slow things down in OWS by doing that here

robbibt: The colour stretches will probably need optimising; as a first pass I chose values that seemed to work OK in the notebook example here: https://gist.github.com/robbibt/285bd20d25b9257dad3b162d965282f5

robbibt:
Have updated the pansharpening formula to use **mean** of all multispectral RGB bands rather than **sum**, as conceptually this should produce the same relative results between pixels, but result in a more sensible scaling.

Have also reverted scaling to the original 0-3000 range.

SpacemanPaul (approved):

PR #1114, opened 2022-11-10
Add S2 C3 OWS layer

jmettes:


robbibt:
Hey @jmettes! Is the idea that we provide both the Sentinel-1 Collection 3 and the Collection 1 OWS layers side-by-side in OWS? My only hesitation about that is that it will make the user experience pretty complicated; most users won't know/care about the difference between the layers, and may be scared off by being offered too much choice. Is it too early to turn off the Sentinel-2 Collection 1 layers?

jmettes:
Yeah, agreed this will make things worse user experience-wise.

There's a bit of a transition period where we have C3 data available, but not ready to decommission C1 data.

Options:
1. Just naively add additional C3 layer (what this PR implements)
2. Add new C3 layer, remove C1 layers (users may still need C1, C3 final forward processing and NRT aren't coming in yet)

So we could just wait a bit longer and implement option 2. Or we can do option 1 now.

robbibt:
One other possible option would be to do what we did for the Provisional data and put them in the "Sentinel-2" subfolder - that way they were still accessible on prod OWS, but not right up front where they could be confusing to users:
image

pindge (approved):

robbibt (changes requested):
One tiny thing apart from the issues above - can we please rename the layers from `"DEA Surface Reflectance (Sentinel-2* MSI) Collection 3` to `"DEA Surface Reflectance (Sentinel-2* MSI, Collection 3)`? That will be more consistent with the format of other products that have extra suffixes in the names

uchchwhash:
> One other possible option would be to do what we did for the Provisional data and put them in the "Sentinel-2" subfolder - that way they were still accessible on prod OWS, but not right up front where they could be confusing to users.

I like this, but as we discussed in the standup, I don't have any strong preference one way or the other.

uchchwhash (approved):

jmettes:
> One other possible option would be to do what we did for the Provisional data and put them in the "Sentinel-2" subfolder - that way they were still accessible on prod OWS, but not right up front where they could be confusing to users

I think this is a good compromise, and my preference, if you're happy with it. This lets users who still need to use more recent C1 data continue (if they can find them! 😆), and at same time not confuse new users.

robbibt:
> > One other possible option would be to do what we did for the Provisional data and put them in the "Sentinel-2" subfolder - that way they were still accessible on prod OWS, but not right up front where they could be confusing to users
> 
> I think this is a good compromise, and my preference, if you're happy with it. This lets users who still need to use more recent C1 data continue (if they can find them! 😆), and at same time not confuse new users.

Just to confirm: would that be adding the Collection 3 layers to the "Sentinel-2" subfolder, or moving the Collection 1 layers to the subfolder? I have a small preference for option 1 (at least until the C3 products are being updated productionally with NRT), but happy with either if anyone has a strong preference 😄

jmettes:
Ok after quick chat with @robbibt, we decided that the main goal is to preserve continuity for users that want up to date NRT. The super-keen users can find the C3 layers in the sub-menu (we can also communicate their existence in correspondence, along with any decommissioning messaging).

image

I am happy to update the terria config again once NRT/forward processing is more current.

robbibt (approved):
Looks great to me, I think this is a good compromise until the main S2 layers are up and running with feature parity with our current layers. Exciting!

PR #1117, opened 2022-11-17
Add load hints to all Collection 3 summary products

robbibt:
A user recently noticed that some of our Collection 3 summary products don't include pre-configued default CRS and resolution details. This negatively affects the experience of using our products, as users recieve error messages about having to supply `output_crs` and `resolution` when loading our Collection 3 products even when this is not actually necessary due to all the products sharing the same Collection 3 grid.

This PR aims to address this issue by making sure all our Collection 3 product definitions contain a `storage` section that includes `crs` and `resolution`, e.g.:
```
storage:
  crs: EPSG:3577
  dimension_order:
  - time
  - y
  - x
  resolution:
    x: 30
    y: -30
  tile_size:
    x: 96000.0
    y: 96000.0
```

I have [taken this section from here](https://github.com/GeoscienceAustralia/dea-config/blob/master/products/inland_water/wofs/ga_ls_wo_fq_cyear_3.odc-product.yaml#LL25-L36C15), as it seemed the most complete and up to date (I'm not sure if `dimension_order` is a compulsory inclusion, but I have included it there for now).

I have also made a number of small other changes:

- Added license details to all our C3 products
- Changed the DEA Landcover C3 definition to use a negative `y=-30` instead of `y=30` for consistency with other products
- Added license for our C2 Landcover and coastal products
- Added correct descriptions for our C2 coastal products

I would really appreciate a careful review for this - I don't have a good grasp of any downstream implications of this, and would rather be safe than sorry. Thanks in advance! 😃 

---

UPDATE: Based on discussion below, have updated this PR to add load hints to all our Collection 3 products:

```
load:
  crs: 'epsg:3577'
  resolution: 
    y: -30
    x: 30
  align:
    y: 0
    x: 0
```


emmaai (commented):
Not sure if tile size should be in pixels or in meters? anybody?

pindge:
the `tile_size` is the `gridspec` https://github.com/opendatacube/datacube-core/blob/ba30625733c606329eb1861302a84ff1d01ebc48/datacube/model/__init__.py#L600-L601

or https://github.com/opendatacube/datacube-core/blob/ba30625733c606329eb1861302a84ff1d01ebc48/datacube/model/__init__.py#L593-L607

so its either `tile_size` or `crs` and `resolution` in `storage` section

pindge:
> Not sure if tile size should be in pixels or in meters? anybody?

as per  the odc doc,

> measurements
List of measurements in this product. The measurement names defined here need to match 1:1 with the measurement key names defined in the dataset-metadata-doc.

we need to check metadata doc?

pindge:
> (I'm not sure if dimension_order is a compulsory inclusion, but I have included it there for now)

Some cursory look, the `dimension_order` is only used for ingestion. Found in this page: https://opendatacube.readthedocs.io/en/latest/installation/ingesting-data/index.html?highlight=dimension_order

for Storage, as per doc is only below

> storage (optional)
Describes some of common storage attributes of all the datasets. While optional, defining this will make product data easier to access and use. This only applies to products that have data arranged on a regular grid, for example ingested products are arranged like that.

crs
Coordinate reference system common to all the datasets in the product. 'EPSG:<code>' or WKT string.

resolution
Resolution of the data of all the datasets in the product specified in projection units. Use latitude, longitude if the projection is geographic and x, y otherwise.

tile_size
Size of the tiles for the data to be stored in specified in projection units. Use latitude and longitude if the projection is geographic, otherwise use x and y.

origin
Coordinates of the bottom-left or top-left corner of the (0,0) tile specified in projection units. If coordinates are for top-left corner, ensure that the latitude or y dimension of tile_size is negative so tile indexes count downward. Use latitude and longitude if the projection is geographic, otherwise use x and y.

robbibt:
> Not sure if tile size should be in pixels or in meters? anybody?

I used metre units because this page indicates it should be in "projection units" - happy to update if I interpreted it wrong

https://opendatacube.readthedocs.io/en/latest/about-core-concepts/products.html

products/land_and_vegetation/lccs/ga_ls_landcover_class_fyear_3.odc-product.yaml
```diff
@@ -63,11 +64,16 @@ measurements:
...
+  - x
+  resolution:
+    x: 30
+    y: -30
   tile_size:
     x: 96000.0
     y: 96000.0
-  resolution:
-    x: 30
-    y: 30
```
tebadi, line 73: y: -30 is the correct value.

robbibt:
> Some cursory look, the `dimension_order` is only used for ingestion.

Thanks Pin, so do you think I should remove it for simplicity if it's no longer used for indexed products? 


> so its either `tile_size` or `crs` and `resolution` in `storage` section and with `crs` and `resolution` as first preference

To clarify: we don't need `tile_size` if we have `resolution`and `crs`?

products/land_and_vegetation/lccs/ga_ls_landcover_class_fyear_3.odc-product.yaml
robbibt: Thanks Toktam!

omad:
My oh my, this ODC documentation is woeful!

While using `storage` like this will work, Kirill implemented a **new** section in the product specification called `load` that is specifically for this. See https://github.com/opendatacube/datacube-core/issues/1011 and https://github.com/opendatacube/datacube-core/pull/1012 .

I think we should *remove* the `storage` section from all summary products, and add a `load` section defining  `crs` and `resolution`.

robbibt:
> @omad While using `storage` like this will work, Kirill implemented a **new** section in the product specification called `load` that is specifically for this. See [opendatacube/datacube-core#1011](https://github.com/opendatacube/datacube-core/issues/1011) and [opendatacube/datacube-core#1012](https://github.com/opendatacube/datacube-core/pull/1012) .
> 
> I think we should _remove_ the `storage` section from all summary products, and add a `load` section defining `crs` and `resolution`.

I'm totally on board with updating the definitions to use `load` instead of `storage` if we're sure this is fully supported - defaulting to loading Collection 3 summary products on the Collection 3 resolution/CRS/alignment makes a lot of sense to me. 

@pindge If we did this, I suspect you'd need to revise [the PR you mad to Core yesterday?](https://github.com/opendatacube/datacube-core/pull/1357) 

Another point for discussion: should we consider adding a recommended `EPSG:3577` CRS and `(-30, 30)` resolution to even our non-summary products too? e.g. `ga_ls8c_ard_3`, `ga_s2am_ard_3`? It would make the user experience much simpler, although at the cost of making users engage mentally with the concept of native pixel grids vs reprojected data etc...

pindge:
> > @omad While using `storage` like this will work, Kirill implemented a **new** section in the product specification called `load` that is specifically for this. See [opendatacube/datacube-core#1011](https://github.com/opendatacube/datacube-core/issues/1011) and [opendatacube/datacube-core#1012](https://github.com/opendatacube/datacube-core/pull/1012) .
> > I think we should _remove_ the `storage` section from all summary products, and add a `load` section defining `crs` and `resolution`.
> 
> I'm totally on board with updating the definitions to use `load` instead of `storage` if we're sure this is fully supported - defaulting to loading Collection 3 summary products on the Collection 3 resolution/CRS/alignment makes a lot of sense to me.
> 
> @pindge If we did this, I suspect you'd need to revise [the PR you made to Core yesterday](https://github.com/opendatacube/datacube-core/pull/1357), so that it can obtain `default_resolution` and `default_crs` from the `load` section too?
> 
> Another point for discussion: should we consider adding a recommended `EPSG:3577` CRS and `(-30, 30)` resolution to even our UTM zone non-summary products too? e.g. `ga_ls8c_ard_3`, `ga_s2am_ard_3`? It would make the user experience much simpler, although at the cost of potentially allowing users to not have to consider the implications of things like resampling etc...

https://github.com/opendatacube/datacube-core/blob/61c25d6745da9b391b33e9b26f9afeeaef059a99/datacube/api/core.py#L852-L860

looks like `load_hints` takes a priority

omad:
> Another point for discussion: should we consider adding a recommended EPSG:3577 CRS and (-30, 30) resolution to even our UTM zone non-summary products too? e.g. ga_ls8c_ard_3, ga_s2am_ard_3? It would make the user experience much simpler, although at the cost of potentially allowing users to not have to consider the implications of things like resampling etc...

Personally, I think this is a good idea. With the assumption that we explain clearly somewhere that it is happening.

The downsides to unintentionally resampling data are no worse than the other [foot guns](https://en.wiktionary.org/wiki/shoot_oneself_in_the_foot) that exist when using DEA data. (eg. data with poor geometric accuracy, not using cloud or water masks, not dilating cloud masks, etc).

robbibt:
> > Another point for discussion: should we consider adding a recommended EPSG:3577 CRS and (-30, 30) resolution to even our UTM zone non-summary products too? e.g. ga_ls8c_ard_3, ga_s2am_ard_3? It would make the user experience much simpler, although at the cost of potentially allowing users to not have to consider the implications of things like resampling etc...
> 
> Personally, I think this is a good idea. With the assumption that we explain clearly somewhere that it is happening.
> 
> The downsides to unintentionally resampling data are no worse than the other [foot guns](https://en.wiktionary.org/wiki/shoot_oneself_in_the_foot) that exist when using DEA data. (eg. data with poor geometric accuracy, not using cloud or water masks, not dilating cloud masks, etc).

Yep, I agree - lots of other big gotchas that would arguably have more impact than this.

As far as I can tell, I don't _think_ a change like this would break anyone's code? It should only affect calls to `dc.load` where no `output_crs` and `resolution` are specified (which should currently produce an error message anyway).

robbibt:
OK, have updated this PR to add load hints to all our Collection 3 products:

```
load:
  crs: 'epsg:3577'
  resolution: 
    y: -30
    x: 30
  align:
    y: 0
    x: 0
```

I'd like to also update our UTM scene-based products to include load hints too, but I'll raise that as a new PR for further discussion once this one is merged as it's a slightly more nuanced issue than this one.

products/land_and_vegetation/lccs/ga_ls_landcover_class_fyear_3.odc-product.yaml
```diff
@@ -63,11 +64,16 @@ measurements:
     dtype: uint8
     nodata: 0
     units: '1'
+
 storage:
   crs: EPSG:3577
+  dimension_order:
```
omad, line 70: Missed one :)

products/land_and_vegetation/lccs/lc_ls_c2.odc-product.yaml
```diff
@@ -1,6 +1,7 @@
 name: ga_ls_landcover_class_cyear_2
 
 description: Geoscience Australia Landsat Land Cover Calendar Year Collection 2.0
+license: CC-BY-4.0
```
omad, line 4: `load` instead of `storage` for this file too.

omad (changes requested):
I think that all of these can be updated to use `load` instead of storage. 

AFAICT, `storage` is a hang-over from when we were *ingesting* data. Which isn't something any of these Products require.

robbibt:
Thanks @pindge! I was going to leave updating the scene-based UTM products for another PR, but as you've made a start on that here we might as well do it all in one place - will push a commit adding the following to our Landsat scene-based products:

```

load:
  crs: 'epsg:3577'
  resolution:
    y: -30
    x: 30
  align:
    y: 0
    x: 0
```

And the following for our Sentinel-2 scene-based products:

```

load:
  crs: 'epsg:3577'
  resolution:
    y: -10
    x: 10
  align:
    y: 0
    x: 0
```

products/land_and_vegetation/lccs/ga_ls_landcover_class_fyear_3.odc-product.yaml
robbibt: Done!

products/land_and_vegetation/lccs/lc_ls_c2.odc-product.yaml
robbibt: Done

mpaget:
Hi, just noting that I'm tracking this for the CSIRO DBs that index DEA products. All good and thank you. I agree that

- consistent load hints will make `dc.load()` easier to use for a majority of products, and
- load hints for UTM products should be considered in the context the products are normally used, with consideration to options and defaults presented in a notebook.

emmaai:
Do we align on corner `(0, 0)` or centre `(-15. 15)`? Is it for pixel or whole tile?

mpaget:
Hi @emmaai, my understanding is that `align` is used where the coordinate variables should be interpreted as aligned to the top left of a pixel `(0, 0)` (default) or to the centre of a pixel `(half-y-resolution, half-x-resolution)`. Broadly:

- top left alignment is the default for geotiff
- centre alignment is the default for netcdf, and as such could be the default for climate-related datasets.

robbibt:
@emmaai My understanding is that `align=(15, 15)` is - only required when loading Landsat data into its native UTM zone CRS and resolution, to ensure that the pixel alignment selected by datacube matches the pixel alignment on file. For any situation where we reproject data (e.g. into `EPSG:3577` like in these load hints), `align` should be set to its default value of `(0, 0)`.

Some notes on thus from @Kirill888 from a few years ago: https://gist.github.com/robbibt/53850ebcabf4dab7311445d9d63e1a97

Kirill888:
@pindge 

> looks like load_hints takes a priority

It does NOT, if there is grid_spec those "hints" are rewritten to what gridspec defines. Both describe a pixel grid, but grid_spec is using meters/degrees, as a result one could have all sorts of nasty rounding issues in there.

products/inland_water/ga_ls_tcw_percentiles_2/ga_ls_tcw_percentiles_2.odc-product.yaml
```diff
@@ -14,29 +15,21 @@ measurements:
...
     y: -25
-  chunking:
-    x: 1000
-    y: 1000
-    time: 1
-  dimension_order: [time, y, x]
+    x: 25
+  align:
+    y: 0
+    x: 0
```
emmaai, line 35: Does the change apply to collection 2 as well? This one is from collection 2.

Kirill888:
@mpaget in `datacube` there is no such thing as `crs="utm"`. In `odc-stac` such thing exists and means "pick appropriate UTM strip for a given query".

emmaai:
thanks @mpaget and @robbibt, guess my fundamental question is if `align` defines how the coordinates are allocated for each pixel GeoTiff/netcdf? i.e., `(0, 0)` means those values of coordinates in the files are top-left corner of pixel? Or alternatively, it defines how we would wanna mark the coordinates in loading no matter how they are stored in files? i.e., by convention, all the coordinates should be top-left corner in GeoTiff, if `align=(-15, 15)`, in the loading results, the coordinates will be shifted to `(-15, 15)`, which is different from the stored coordinates values.

products/inland_water/ga_ls_tcw_percentiles_2/ga_ls_tcw_percentiles_2.odc-product.yaml
robbibt: Yeah, the scope has crept to include Collection 2 products since I originally started the PR - personally I'd be fine with restricting this to Collection 3 products only, but I can see how it would be useful to have a consistent approach across all our products. Thoughts, @pindge @omad?

Kirill888:
> thanks @mpaget and @robbibt, guess my fundamental question is if align defines how the coordinates are allocated for each pixel GeoTiff/netcdf? i.e., (0, 0) means those values of coordinates in the files are top-left corner of pixel? Or alternatively, it defines how we would wanna mark the coordinates in loading no matter how they are stored in files? i.e., by convention, all the coordinates should be top-left corner in GeoTiff, if align=(-15, 15), in the loading results, the coordinates will be shifted to (-15, 15), which is different from the stored coordinates values.

load hints are all about load time configuration, completely independent of data, `storage` section should probably faithfully describe tiling regime that was used during ingestion. This is defining output pixel grid dc.load should create in the absence of configuration coming from the user. This does not introduce bias by itself, only if you use wrong align WHILE ALSO using `nearest` reprojection WHILE ALSO loading in the native projection and resolution.

Kirill888:
remember that there is not "native" load in datacube. Nor is it possible in the general case, only something like "mostly native" is possible in the general case. When merging different images whether mosaic or temporal stac, everything being native is just not possible in the general case, only when data has all the compatible grids.

There is "best effort native" implementation in `odc-stac` though.

pindge (approved):

PR #1118, opened 2022-11-22
rename lccs cyear file

pindge:


robbibt:
Looks great Pin. Should we rename  the entire folder to "landcover"? LCCS is the old name I think (ping @tebadi)

tebadi:
> Looks great Pin. Should we rename the entire folder to "landcover"? LCCS is the old name I think (ping @tebadi)

Good Suggestion Robbie! We no longer refer to landcover as LCCS as you said.

pindge:
@tebadi @robbibt commit pushed

robbibt (approved):
Thanks Pin!

PR #1121, opened 2022-11-29
Add load hints and license to remaining products

robbibt:
This PR adds license and load hints to remaining products with missing data here:
https://explorer.dev.dea.ga.gov.au/audit/product-overview

Products I couldn't find:
[s2_l2a](https://explorer.dev.dea.ga.gov.au/products/s2_l2a)
[ga_ls_wo_summary](https://explorer.dev.dea.ga.gov.au/products/ga_ls_wo_summary)

Products that show no default CRS and resolution but seem to have load hints already correctly set (maybe you missed updating this?)
[ga_s2_nrt_c3_wo](https://explorer.dev.dea.ga.gov.au/products/ga_s2_nrt_c3_wo)


pindge:
I guess the best we can do is to save a copy of this in wofs? https://explorer.dev.dea.ga.gov.au/products/ga_ls_wo_summary.odc-product.yaml

robbibt:
@pindge I'm pretty happy with this PR, even though we haven't worked out what the `ga_ls_wo_summary` product is. I think we could probably merge it for now and update that product later if we really want to? (although I don't think it will be necessary as I'm pretty sure it's deprecated)

robbibt:
OK, from Norman's message, I don't think we need to worry about that product any more 🙂

pindge (approved):

PR #1125, opened 2022-12-05
add decommission notice to c1 s2 layers

pindge:
- add formating to readme
- add decommission msg to c1 s2 layers for both dev and prod


CEKrause (approved):
Thanks @pindge. That all looks correct to me.

PR #1131, opened 2023-01-22
remove sentinel collection 1 and add collection 3 into terriacube in dev

JM-GA:
I am not too familiar with how terriacube works - please let me know if something needs changing. I have removed sentinel 2 collection 1 products from the 'DEA Production' section and replaced with collection 3 leaving the same ids.


jmettes:
I've updated the Production section of the dev terria config to include the high level S2 surface reflectance catalogue item. This is present in the actual DEA Maps production catalogue too.

jmettes (approved):

PR #1132, opened 2023-01-23
Allow Caching of OWS Responses

omad:
At the moment, approximately none (0.44% in the last 30 days) of our OWS traffic is being cached by CloudFront. This is a problem for two reasons. 1: poor performance for OWS Users. 2: Increased load on OWS Application Servers.

This change allows GetCapabilities requests to be cached for 1 hour, and allows GetMaps/GetTile requests to be cached for 1 hour.

Previously, the only requests being cached are those for GetMap/GetTile requests that combine 4 or more datasets.

### DIFF
--- a/prod/services/wms/ows_refactored/ows_reslim_cfg.py
+++ b/prod/services/wms/ows_refactored/ows_reslim_cfg.py
@@ -3,6 +3,10 @@
 
 dataset_cache_rules = [
     {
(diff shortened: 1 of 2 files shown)


jmettes (approved):

jmettes:
gimme the CACHEEE
image

PR #1133, opened 2023-01-23
switch to new s2 c3 nrt product

JM-GA:
switch products in the alchemist product specification

### DIFF
--- a/dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml
+++ b/dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml
@@ -52,7 +52,6 @@ output:
 
   properties:
     odc:product: ga_s2_wo_provisional_3
-    dea:dataset_maturity: interim
     dea:product_maturity: provisional
     eo:instrument: MSI
     odc:file_format: GeoTIFF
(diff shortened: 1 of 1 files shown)


jmettes:
I think we may want `dea:dataset_maturity: interim` to be removed? I believe [alchemist was recently updated](https://github.com/opendatacube/datacube-alchemist/pull/155) to source the maturity from the dataset rather than the config, so it can be dynamic.

JM-GA:
@jmettes - yeah I wasn't too sure about this one - I was thinking that the maturity flag was for the metadata for the output rather than filtering by maturity for the input - so left it as is to keep the data consistent - but also happy to change

JM-GA:
@jmettes - oh sorry I see what you mean - so better to remove the maturity line completely?

jmettes:
We might be able to merge [s2_nrt_wo.alchemist.yaml](https://github.com/GeoscienceAustralia/dea-config/blob/943ae5a9f45aa7f39438a8cd4e69ddfee581c3a3/dev/services/alchemist/s2_nrt_wo/s2_nrt_wo.alchemist.yaml) and [s2_c3_nrt_wo_provisional.alchemist.yaml](https://github.com/GeoscienceAustralia/dea-config/blob/943ae5a9f45aa7f39438a8cd4e69ddfee581c3a3/dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml] in that case? Since the only real difference is the dataset maturity

JM-GA:
@jmettes  - [s2_nrt_wo.alchemist.yaml](https://github.com/GeoscienceAustralia/dea-config/blob/943ae5a9f45aa7f39438a8cd4e69ddfee581c3a3/dev/services/alchemist/s2_nrt_wo/s2_nrt_wo.alchemist.yaml) is based on collection 1 input data. I am waiting on Norman to confirm if I can decommission it

jmettes (approved):

PR #1136, opened 2023-01-25
Add new combined S2 and Landsat layer to Terriacube

robbibt:
This PR adds the new combined Sentinel-2 and Landsat OWS layer to Terriacube. I've added an "OWS" suffix to this layer, and an "GSKY" suffix to the old layer. 

For now I've left the FeatureInfoTemplate to use the Landsat feature info template, but this currently only includes the RGB Landsat bands that are currently included in the product. I expect this will work better once we add the other bands, but we should remember to have a look at it later on.

image

### DIFF
--- a/dev/terria/terria-cube-v8.json
+++ b/dev/terria/terria-cube-v8.json
@@ -39,7 +39,7 @@
                             "leafletUpdateInterval": 750,
                             "chartType": "momentPoints",
                             "featureInfoTemplate": {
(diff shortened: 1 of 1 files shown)


dev/terria/terria-cube-v8.json
```diff
@@ -26,6 +26,34 @@
...
...a.0.bands.nbart_red}}</td></tr>  <tr><td><b>Near Infrared (NIR) - 840</b></td><td>{{data.0.bands.nbart_nir}}</td></tr>  <tr><td><b>Shortwave Infrared (SWIR) - 1650</b></td><td>{{data.0.bands.nbart_swir_1}}</td></tr>  <tr><td><b>Shortwave Infrared (SWIR) - 2220</b></td><td>{{data.0.bands.nbart_swir_2}}</td></tr>  <tr><td><b>Normalised Difference Vegetation Index</b></td><td>{{data.0.band_derived.ndvi}}</td></tr>  <tr><td><b>Normalised Difference Water Index</b></td><td>{{data.0.band_derived.ndwi}}</td></tr><tr><td><b>Modified Normalised Difference Water Index</b></td><td>{{data.0.band_derived.mndwi}}</td></tr><tr><td><b>Normalised Burn Ratio</b></td><td>{{data.0.band_derived.nbr}}</td></tr></table><p><chart id='{{x}}{{y}}{{time}}' title='NBART at {{x}},{{y}}' column-units='Wavelength (nm), Reflectance x 10k' preview-x-label='NBART Reflectance'>nm,NBART Reflectance\n490,{{data.0.bands.nbart_blue}}\n560,{{data.0.bands.nbart_green}}\n660,{{data.0.bands.nbart_red}}\n840,{{data.0.bands.nbart_nir}}\n1650,{{data.0.bands.nbart_swir1}}\n2220,{{data.0.bands.nbart_swir2}}</chart></p><p><b>Imagery available for dates:</b>{{#data_available_for_dates}}<br/>{{.}}{{/data_available_for_dates}}</p>",
```
SpacemanPaul, line 42: The feature info template includes the infra-red bands that are explicitly not included in the OWS combined layer.

SpacemanPaul (approved):
You might want to drop the infrared bands from the feature info template.  We will have to rewrite it when the infrared bands get in anyway, as they will most likely have different names in the combined layer.

PR #1134, opened 2023-01-24
Proof-of-concept combined landsat-sentinel-2 layer.

SpacemanPaul:
Proof-of-concept combined landsat-sentinel-2 layer.  

Visible bands only, no masking.  Dev environment only.

PR #1137, opened 2023-01-29
Add experimental (RGB bands only) S2/LS combined layer to prod.

SpacemanPaul:
Prototype S2/LS combined layer, currently restricted to visible bands only (RGB) due a lack of common band aliases for the mappable infrared bands.

Added to dev in #1134.

Note I have NOT added the new prod layer to the Terria catalog.  This is still an experimental layer - testing on the prod database because S2C3 data in dev is still very patchy.

### DIFF
--- a/prod/services/wms/inventory.json
+++ b/prod/services/wms/inventory.json
@@ -1,10 +1,10 @@
 {
-    "total_layers_count": 69,
+    "total_layers_count": 65,
     "layers": [
          {
             "layer": "s2_ls_combined",
-            "product": [ 
-                "ga_s2am_ard_3", 
+            "product": [
+                "ga_s2am_ard_3",
                 "ga_s2bm_ard_3",
                 "ga_ls5t_ard_3",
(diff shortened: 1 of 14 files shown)


prod/services/wms/inventory.json
```diff
@@ -1,6 +1,24 @@
 {
-    "total_layers_count": 64,
+    "total_layers_count": 69,
```
pindge, line 2: its only adding `1` layer, the count should be `64` + `1` = 65

pindge (changes requested):

robbibt (approved):
This looks great to me @SpacemanPaul in terms of band names, descriptions, naming etc - will review the WMS itself in more detail once it's live in OWS (e.g. compare it directly against the existing GSKY layer)

prod/services/wms/inventory.json
robbibt: This looks like it has been fixed

PR #1138, opened 2023-01-30
Test out OWS GetMap/GetTile caching in DEA Dev OWS

omad:
My attempt last week to enable caching failed, see https://github.com/opendatacube/datacube-ows/issues/921

I think I know what went wrong, there was no `max_datasets` set for the layers, and the caching logic decided not to cache anything.

I'm not 100% sure what a reasonable `max_datasets` is, so I've copied the value of 24 from the WCS config.

This change is against DEA Dev OWS, if it's successful, we can make it for Prod too.

### DIFF
--- a/dev/services/wms/ows_refactored/ows_root_cfg.py
+++ b/dev/services/wms/ows_refactored/ows_root_cfg.py
@@ -92,6 +92,8 @@ ows_cfg = {
         "s3_aws_zone": "ap-southeast-2",
         "max_width": 512,
         "max_height": 512,
+        # Allow the WMS/WMTS GetCapabilities responses to be cached for 1 hour
+        "caps_cache_maxage": 3600,
     },  # END OF wms SECTION
     "wmts": {
         # Config for WMTS service, for all products/layers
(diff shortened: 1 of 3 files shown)


SpacemanPaul:
That should work - although it does mean that all non-empty tiles will be cached for at least one hour.

The risk with that is that the cache gets filled up with single-dataset tiles which can be served quickly on the fly, pushing out the more expensive multi-dataset tiles which take much longer to serve.  But probably worth testing.

We don't usually apply max_dataset limits to WMS/WMTS layers.  24 should be fine though for most products though.

omad:
CloudFlare is mostly a black box, but the documentation indicates that tiles don't get "pushed out". It can respond *much* faster than OWS can.

> We don't usually apply max_dataset limits to WMS/WMTS layers. 24 should be fine though for most products though.

See https://github.com/opendatacube/datacube-ows/issues/921, if `max_datasets` isn't set, it's impossible to cache Map/Tile requests. I think it's a bug

dev/services/wms/ows_refactored/ows_root_cfg.py
```diff
@@ -92,6 +92,8 @@
         "s3_aws_zone": "ap-southeast-2",
         "max_width": 512,
         "max_height": 512,
+        # Allow the WMS/WMTS GetCapabilities responses to be cached for 1 hour
+        "caps_cache_maxage": 3600,
```
SpacemanPaul, line 96: In general I would prefer something like `60 * 60`  - but 3600 should be familiar enough as the number of seconds in an hour.

SpacemanPaul (approved):
I agree there is an OWS bug to require max_datasets.

This should work though I think.

PR #1151, opened 2023-02-01
Add common aliases for S2/LS bands

jeremyh:
This adds the following "common" alias names to S2 and L1 ARD products, so that a single band name can be used consistently across them.

| alias              | mapped to landsat     | mapped to sentinel    |
|--------------------|-------------|-------------|
| nbar_common_nir    | nbar_nir    | nbar_nir_1  |
| nbar_common_swir_1 | nbar_swir_1 | nbar_swir_2 |
| nbar_common_swir_2 | nbar_swir_2 | nbar_swir_3 |

... as well as the equivalent three aliases for `nbart`

These match the definitions mentioned in our S2/LS table here: https://docs.dea.ga.gov.au/notebooks/Beginners_guide/02_DEA.html

---

(I'm open to alternative names for these @SpacemanPaul @robbibt )


robbibt:
Hey @jeremyh, I like the band names! Can we also get this done for Landsat 9? I think it's included in the combined product, so good to have it consistent too
https://github.com/GeoscienceAustralia/dea-config/blob/51da611071c060ac012d81f4f394c48f9d7fede1/products/baseline_satellite_data/c3/ga_ls9c_ard_3.odc-product.yaml

robbibt:
Also, I'm unfamiliar with this `ga_ls8c_ard_c2_3` product... it has "c2" in the name but "3" (for Collection 3) in the suffix... we shouldn't have any Landsat C2 data left anymore, so can it be removed completely?
https://github.com/GeoscienceAustralia/dea-config/blob/51da611071c060ac012d81f4f394c48f9d7fede1/products/baseline_satellite_data/c3/ga_ls8c_ard_c2_3.odc-product.yaml

robbibt:
Hey @jeremyh, let me know when this is ready for review! So far it looks great to me

jeremyh:
> I'm unfamiliar with this `ga_ls8c_ard_c2_3` product... it has "c2" in the name but "3" (for Collection 3) in the suffix... we shouldn't have any Landsat C2 data left anymore, so can it be removed completely?

The C2 seems to be referring to USGS's Collection 2, not DEA's Collection 2. That ever-present confusion of collection numbering :)

robbibt (approved):
This looks great to me, and all the bands match their correct common aliases.

I would probably think we should remove the `ga_ls8c_ard_c2_3` product from `dea-config` entirely as I don't think we use it for anything now (it's definitely not used in OWS), and it just makes things confusing. But that might be a job for another PR!

PR #1157, opened 2023-02-13
IR bands for LS/S2 combined layer.

SpacemanPaul:
For prod.  (Dev is not appropriate for testing as it has minimal S2C3 data - combined layer has not yet been added to Terria config, so should be invisible to users.)

Adds common NIR and SWIR bands, and styles that used them.  Based closely on existing LS5 config.

I've removed references to exact central wavelengths as they vary slightly from detector to detector and platform to platform.  I've given rounded off "approximate" wavelengths instead.

### DIFF
--- a/prod/services/wms/inventory.json
+++ b/prod/services/wms/inventory.json
@@ -11,12 +11,21 @@
                 "ga_ls8c_ard_3",
                 "ga_ls9c_ard_3"
             ],
-            "styles_count": 4,
+            "styles_count": 13,
             "styles_list": [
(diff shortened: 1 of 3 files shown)


robbibt (approved):
As far as I can tell from skimming the code, this looks really nice! I agree with the choice to remove/simplify the precise wavelength info on this layer - makes more sense to have them generalised.

Excited to have a combined Fmask here too!

Will do a more detailed test of the layer when it's available in the WMS.

PR #1160, opened 2023-02-13
Update combined S2 Landsat layer in Terriacube

robbibt:
This PR updates the FeatureInfoTemplate section of our Terriacube config so that it will work with the recently added bands on the combined Sentinel-2 and Landsat layer.

This does _not_ update DEA Maps, only Terriacube.


SpacemanPaul (approved):
LGTM

JM-GA (approved):

PR #1163, opened 2023-02-15
Add new Github Actions workflow to post preview link to modified Terria configs

robbibt:
When someone makes edits to our Terria JSON configs (e.g. `terria-cube-v8.json` or `dea-maps-v8.json`), it can be difficult to easily tell how these changes look in Terria itself. Currently users have to download the config to their PC, create a clean Terria instance by appending #clean to the Terria URL, then drag and drop the updated config file into Terria to review their changes.

This PR attempts to simplify this by creating a new Github Actions bot that does the following:

1. If a PR is raised to `master` and modifies any of the Terria config files inside `dev/terria/...`, then:
2. Creates links to Terria pointing to the raw configs stored on the relevant Github feature branch (e.g. `terria_link_preview2` for this PR)
3. Posts these URLs as a comment on the pull request.

That way, users can easily click the links and inspect the config changes on the map to help them review the PR!

image

Feedback and suggestions welcome! This was just a fun activity to do while AWS is down this afternoon (inspired by a Teams chat about the recent OWS Sentinel-2 + Landsat combined layer config changes).


.github/workflows/terria_preview.yaml
```diff
@@ -0,0 +1,36 @@
+name: Terria catalog preview
+
+on:
+  push:
```
omad, line 4: I don't think this `push` section is used, or necessary. It looks like this is just for adding comments to PRs

omad (approved):
Awesome Robbi! I really love this!

.github/workflows/terria_preview.yaml
robbibt: Yeah, I think that makes sense!

robbibt:
Hey @JM-GA @SpacemanPaul  - I'll merge this for now, but we can test it out when we come to removing the GSKY combined layer and adding the combined OWS layer to DEA Maps!

JM-GA:
Sounds great! thanks @robbibt!

PR #1164, opened 2023-02-20
Decommission ls nrt prov

JM-GA:
remove landsat nrt c3 provisional web services


dev/terria/dea-maps-v8.json
```diff
@@ -6,24 +6,6 @@
                     "type": "group",
                     "name": "Baseline satellite data",
                     "members": [
-                        {
-                            "type": "wms",
-                            "name": "DEA Surface Reflectance (Landsat and Sentinel-2)",
-                            "url": "https://gsky.nci.org.au/ows/dea",
```
jmettes, line 12: So this is all good to get replaced (this is the production DEA Maps config, despite it being in dev folder...)? Is there a combined one in already?

JM-GA: @jmettes - yeah CCB approved for removal. As its obsolete now - we will remove it as its no longer being updates and Robbi going to put in PR to put new one in using the same ID. New composite layer is in dev, tested and ready for prod deployment.

jmettes (approved):

PR #1166, opened 2023-02-22
Update terria-cube-v8.json

vnewey:
Update to the the DEA Coastlines entry under the DEA Production section.


robbibt (approved):
Thanks @vnewey! You're right, this was way out of date!

PR #1169, opened 2023-02-23
Update s2_c3_nrt_wo_provisional.alchemist.yaml

JM-GA:
add dataset maturity

### DIFF
--- a/dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml
+++ b/dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml
@@ -52,6 +52,7 @@ output:
 
   properties:
     odc:product: ga_s2_wo_provisional_3
+    dea:dataset_maturity: interim
     dea:product_maturity: provisional
     eo:instrument: MSI
     odc:file_format: GeoTIFF


pindge (approved):

dev/services/alchemist/s2_nrt_wo/s2_c3_nrt_wo_provisional.alchemist.yaml
```diff
@@ -52,6 +52,7 @@ output:
 
   properties:
     odc:product: ga_s2_wo_provisional_3
+    dea:dataset_maturity: interim
```
jmettes, line 55: To be honest, I'm not sure if this field is needed anymore, since alchemist was update to carry over the maturity from input datasets into the outputs. And I don't think S2 has all that many interim datasets? Would be mostly NRT and final

(Also just in case that was the interpretation, interim != provisional.)

JM-GA: unfortunately it is needed as the new product doesn't correctly contain the maturity flag in the input dataset. Essentially all scenes were failing as dataset maturity flag was missing (guessing SNS message does not contain maturity flag) - once I added it in and reset the pods it started processing again. Re interim - that is correct (it should be NRT) but the products were already incorrectly set up and have been interim since creation - both BA and WO are incorrect but I don't want to be making any changes at this stage as stakeholders already using it.

PR #1171, opened 2023-02-26
Remove original DEA Maps catalogue now that it is being read from prod directory

robbibt:
TerriaJS has confirmed that they are now reading the DEA Maps TerriaJS catalogue from the new location in `prod/terria`:
image

This means we can now delete the original catalogue from its location in `dev/terria`.


jmettes (approved):
Wow it's finally gone... thanks you!!

PR #1178, opened 2023-03-14
Use year format for Tasseled Cap Indices Calendar Year

robbibt:
The DEA Tasseled Cap Indices Percentiles Calendar Year (Landsat) product currently uses full date strings for each timestep:

image

This PR formats these dates as years for consistency with all our other annual products:

image


jmettes (approved):

BexDunn (approved):
very useful change, thanks

PR #1179, opened 2023-03-16
ga-s2-wo-3 add ows and into terria-cube

JM-GA:
Adding new web service in production, and adding it to production section in terria cube for initial testing.


pindge (approved):

SpacemanPaul (approved):
Looks good to me.  

Would be nice if we could have definitive, provisional and NRT in the one product like we do with ARD. I assume there was a decision made at some point that this was too difficult.

JM-GA:
@SpacemanPaul the 'provisional' naming relates to the maturity of the product rather than that of the data. (e.g. nrt, interim, and definitive). For this one - the product is a 'not validated' prototype which is why it is labelled provisional

PR #1184, opened 2023-03-23
Add 48 km and 32 km coastal grids to dev Terria

robbibt:
I often find myself wanting to quickly inspect the 48 km and 32 km coastal grids for DEA Coastlines or DEA Intertidal (e.g. to choose what tile to run). This adds those grids to dev Terria so it's easier to inspect them using Terria! 😃 

Should appear under "DEA Development > Other > DEA Shapes"

image


vnewey (approved):
Great idea Robbi!

PR #1187, opened 2023-05-19
adding bc to dev terria catalogue and combining bc style

erin-telfer:
This is a draft to test terria catalogue and combining measurements from same product into one style.


erin-telfer:
To do in next PR
- add descriptive text, e.g. within the abstract
- add combined burncube legend (I will make this with mpl)

SpacemanPaul (approved):

PR #1191, opened 2023-06-02
Prototype Australia Bathymetry product

omad:
This product can be tested in the [Dev Sandbox](https://sandbox.dev.dea.ga.gov.au/) right now and is visible in [ga_multi_ausbath_0: Dev Explorer](https://explorer.dev.dea.ga.gov.au/products/ga_multi_ausbath_0).

Tasks

- [X] Link back to the ecat documentation
- [X] Provide a thumbnail
- [X] tweak some names
- [X] Make sure the data is actually loadable!


robbibt:
As discussed on teams, the "load" section isn't working as intended - may need to swap to "latitude", "longitude" instead of "x" and "x".

robbibt:
Everything looks great here from a desktop review; once this is in dev I'll test it out then approve this PR!

omad:
Thanks @robbibt , I've updated the Product and Dataset in dev, can you please test it out in a notebook to see if the data loading defaults work now.

Also see, thumbnail in explorer!
image

robbibt (changes requested):
This looks fantastic @omad! Have confirmed the data loads perfectly on dev Sandbox, including the fixed "load_hints" (they work fine now) and masking nodata.

The only (minor) thing amiss seems to be the product's location in Explorer: it seems to be in "Hazards" when "Other products" would probably be a better fit. Would this be an easy fix?

https://explorer.dev.dea.ga.gov.au/products
image

omad:
Thanks Robbi!

Yep, we can move things around in Explorer. At the moment that's configured in explorer, not as part of the product defined here.

robbibt (approved):
@omad In that case, awesome! Ready for CCB I think

PR #1193, opened 2023-06-13
add water obs sentinel nrt webservice to dea maps

JM-GA:


robbibt (approved):
Looks great to me - exciting!

PR #1194, opened 2023-06-20
update burncube ows to new product

erin-telfer:
We have just indexed the latest version of burncube to odc. I have updated the ows, and terria catalogue to point at new product + minor wording edit.

### DIFF
--- a/dev/services/wms/inventory.json
+++ b/dev/services/wms/inventory.json
@@ -984,9 +984,9 @@
             ]
         },
         {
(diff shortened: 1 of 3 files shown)


SpacemanPaul (approved):
Looks good to me.

I would just mention as an aside that there is no requirement for the OWS layer name to match the ODC product name.  This a natural thing to do when an OWS layer directly maps to a single ODC product and is fairly common practice - but this (already short) PR could have been even shorter if you'd given the OWS layer a more generic name to begin with.  :)

PR #1198, opened 2023-07-12
Add s2 bc cyear ows

erin-telfer:


erin-telfer:
Looks good :)

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
```diff
@@ -0,0 +1,118 @@
...
+        {"value": 0.0, "color": "#FFFFFF", "alpha": 0.0},
+        {"value": 1, "color": "#F79646"},
+        {"value": 2, "color": "#C0504D"},
+    ],
+    "legend": {
+        "url": "https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/projects/burn_cube/legend_combined.png",
+    },
+}
+
+style_severe_burncube = {
```
SpacemanPaul, line 25: You are mixing two diffferent styling models here.

Colour ramp styles: provide (an `index_function` or an `index expression`) AND a `color_ramp`, as in the combined style above.

Colour Map styles: provide a `value_map` section.

I think you just need to remove the `index_function` here.

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
```diff
@@ -0,0 +1,118 @@
...
+    },
+    "legend": {},
+}
+
+style_moderate_burncube = {
+    "name": "moderate_burncube",
+    "title": "BurnCube: Moderate",
+    "abstract": "burnt area",
+    "needed_bands": ["wofsmoderate"],
+    "index_function": {
```
SpacemanPaul, line 55: As above, remove the index function, it is not needed.

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
```diff
@@ -0,0 +1,118 @@
...
+Roberts, Dale & Mueller, Norman & McIntyre, Alexis. (2017). High-Dimensional Pixel Composites From Earth Observation Time Series. IEEE Transactions on Geoscience and Remote Sensing. PP. 1-11. 10.1109/TGRS.2017.2723896.
+
+Draft CMI https://cmi.ga.gov.au/data-products/dea/841/draft...-burncube-historic-burnt-area
+
+For service status information, see https://status.dea.ga.gov.au""",
+            "product_name": "ga_s2_ard_bc_cyear_3",
+            "bands": bands_in_burncube,
+            "resource_limits": reslim_standard,
+            "native_crs": "EPSG:3577",
+            "native_resolution": [30, -30],
```
SpacemanPaul, line 103: Isn't this a Sentinel-2 layer?  Shouldn't the resolution be `[10, -10]`?

SpacemanPaul (changes requested):
A few minor changes needed, but mostly looks good - see inline comments.

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
Matt-dea: Good catch Paul, but the Sentinel burncube product was derived at a 30x30m resolution to save on processing. Considering 30m is the native res, I'm guessing it should be left as is? Happy to change if not the case.

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
Matt-dea: Thanks for the explanation Paul, I have removed the index functions from the Colour Map styles.

dev/services/wms/ows_refactored/hazards/burntarea/ows_s2_burncube_cyear_cfg.py
Matt-dea: Thanks Paul, actioned.

SpacemanPaul (approved):

PR #1200, opened 2023-07-19
adding LS FY burncube OWS

eloise-b:


erin-telfer:
Nearly there :)
1. add the layer to the ows config - https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/services/wms/ows_refactored/hazards/ows_category_root_cfg.py
2. remove the group heading above the member (our product added to teriia catalogue) and add as a member to the previous group https://github.com/GeoscienceAustralia/dea-config/blob/master/dev/terria/terria-cube-v8.json. I can explain in person :)

erin-telfer:
Can we remove the product from the Terria catalogue altogether? Once the product is on ows then it will be much easier to check bugs?

erin-telfer (approved):

PR #1201, opened 2023-07-20
add other BC to Terria catalogue

erin-telfer:
Ls fyear
S2 cyear


SpacemanPaul (approved):
I don't know Terria config as well as I know OWS, but LGTM.

PR #1202, opened 2023-07-24
update ows hazard structure

erin-telfer:
fixing error with the way the ows catalogue structure was set up for hazards.


SpacemanPaul (approved):
LGTM - but there is a lot of repetition (4 very similar layer).

You might like to look into [configuration inheritance](https://datacube-ows.readthedocs.io/en/latest/configuration.html#configuration-inheritance)  which allows you to create one complete layer, then have other layers inheriting from that layer and just over-riding the elements that are different.

PR #1206, opened 2023-08-31
Deprecate BA product

erin-telfer:
Deprecate the Sentinel 2 Collection 3 NRT Burnt Area Product as CCB No: 20230809.1
Includes:
- removal from DEA Maps
- renaming product in ows and adding note to abstract
- renaming product in Terria


SpacemanPaul (approved):
LGTM

PR #1210, opened 2023-10-17
Fix broken Himawari mosaic layer

robbibt:
Update "8" to "9" in product name as per advice from Arek


omad (approved):
Looks good in the DEA Maps preview.

image

PR #1211, opened 2023-10-23
Update DEA Coastlines charts in Terria

robbibt:
Realised that DEA Coastlines was missing the latest 2022 data in the charts that appear when you click on the map.

This PR adds that data to the chart, and updates some of the wording to describe shorelines more generically (avoiding us having to update the wording every year). E.g. from "2021" to "the most recent shoreline".


erialC-P (approved):
Looks good to go! I like the rewording :)

PR #1220, opened 2023-12-13
Update CMI links and fix broken Coastlines field

robbibt:
This PR:

- Removes any CMI references from our Terria catalogues (affecting DEA Wetlands Insight Tool, DEA Coastlines and DEA Intertidal), replacing them with Knowledge Hub links
- Fixes a bug with a misnamed DEA Coastlines field


CEKrause (approved):
Thanks Robbi. The links are correct. Thanks for updating these.

PR #1222, opened 2024-01-03
Add approved Waterbodies 3.0 text to terriamap

BexDunn:
Updating the waterbodies information to match text agreed upon by contributing folk

Working on:
- improving display of text and information

via @robbibt:

1. There's currently a lot of text on the pop-ups, which made it difficult for me to mentally take in the really nice new customised waterbody data. I would suggest moving the new customised text up above the graph so it jumps out nice and obviously to the user at first glance (see crappy annotation below) 
 2. In Coastlines, we bolded our customised text fields to make them really obvious - could look nice for this too! 
 3. I'd be tempted to reduce the font size of some of the "fine print" text, so it's a little less visually dominant


robbibt (changes requested):
Hi all, this looks fantastic - technically everything looks on point and I've been able to load and use the data nicely via the preview Terria links.

Some tiny suggestions:
1) I think I'd still like to see the waterbody header moved up to the top rather than in the middle next to the graph (and ideally in a slightly larger font to make it more visible).
2) The DEA link here doesn't work - I think it would also be better to point users to the Knowledge Hub instead for "detailed information", rather than back to the website.

image

Other than that, I think it looks great! Will we be removing the older layers soon too?

BexDunn:
Should be ready to go @robbibt  @LaurenSchenk1

robbibt (approved):
Looms great!

PR #1227, opened 2024-02-21
test waterbodies group in prod catalog

BexDunn:
test waterbodies group in prod catalog.
Text is approved, we will be testing that the prod geoserver works correctly.


BexDunn:
> test waterbodies group in prod catalog. Text is approved, we will be testing that the prod geoserver works correctly.

I think we're all good to go - this all works in the DEA Production Terria link, and we're using the prod geoserver. You might notice that the text in the geoserver is coming through as v2 - we have a ticket in for that one with Hellfish.

robbibt:
Hey @BexDunn , this looks great in the prod section of Terria Cube. Is that all you wanted to do for this PR?

(If you want it in DEA Maps itself, you'll have to add it to the DEA Maps catalogue here: https://github.com/GeoscienceAustralia/dea-config/blob/master/prod/terria/dea-maps-v8.json)

robbibt (approved):
Looks great! Onwards to DEA Maps! 🚀

PR #1231, opened 2024-02-29
add ls9 wo

emmaai:
As title. It's prepared for the time when pipeline is triggered once ls9 datasets are pushed into sqs.


JM-GA (approved):

omad (approved):
I'm not sure if Alchemist is pointing to this file directly on GitHub. I think it might be, in which case all we need to do to deploy is merge this PR.

emmaai:
can config Achemist points to this file path on master branch.

PR #1233, opened 2024-03-01
Temporarily restore current GeoMAD product definitions

robbibt:


emmaai (approved):

BexDunn (approved):
Happy with this if @emmaai is :)

PR #1237, opened 2024-03-04
geoMAD update bands to include nbart

JM-GA:
Update band names to include 'nbart'

Once the existing web service is decommissioned - I will put in a PR to remove 'band_c3_ls_cfg.py' and 'style_c3_ls_nbart_cfg.py' which reference existing band names


SpacemanPaul (commented):
If you add the old names as aliases e.g.:

`"nbart_red": ["nbart_red", "red"],`

Then you should be able to re-use the old styles.

JM-GA:
sounds good - I will update PR

SpacemanPaul (commented):
In OWS band dictionaries:

1. The "key" is the "canonical" band name - it is the name that is used internally when interacting with the ODC.  It MUST be a band name/alias listed in the ODC metadata. 

2. The values are aliases that may be used within OWS style definitions.  They may be aliases listed in the ODC metadata, or maybe unique to OWS.

3. The first value in the alias list has special significance - it is the "display name" used for the band, e.g. in GetFeatureInfo responses.

So I think what you want is something more like:

I think what you want is more like:

`"nbart_red": ["nbart_red", "red"],`

JM-GA:
Thanks @SpacemanPaul for the explanation re ows band dictionaries - that makes sense - should be good now

SpacemanPaul (approved):

PR #1239, opened 2024-03-05
Updating all Knowledge Hub links in entire repo

benji-glitsos-ga:
In the entire repo, updated all Knowledge Hub links to point to the new domain: "docs.dea.ga.gov.au" => "knowledge.dea.ga.gov.au"


robbibt (approved):
Not sure if you wanted a review on this @benji-glitsos-ga , but it looks good to me!

benji-glitsos-ga:
Thanks @robbibt . One check is failing 'OWS DEV CFG Test' - is that normal?

robbibt:
> Thanks @robbibt . One check is failing 'OWS DEV CFG Test' - is that normal?

Yep, it's been failing for a few weeks now - it sometimes works if you re-run the tests (have tried that now), but we've been ignoring it recently even if it fails.

PR #1242, opened 2024-03-14
Densify `abares_clum_2020` STAC geometry

robbibt:
While running DEA Intertidal workflows, we realised that the `abares_clum_2020` layer was failing to load in southern Tasmania, despite the GeoTIFF covering all of Australia. 

Comparing the raster against the STAC metadata, we can see that the metadata geometry fails to cover all of the raster: 
image

This seems to have occured due to the dataset's EPSG:3577 bounding box geometry being naively transformed to EPSG:4326 coordinates without accounting for non-linear transformation. To get a better geometry, I've re-run `rio stac` with the `--densify-geom 10` param to densify the geometry before converting to EPSG:4326:
```
rio stac https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/abares_clum_2020/clum_50m1220m.tiff --densify-geom 10 | jq > abares_clum_2020.stac-item.json
```

This looks much better (red):
image


omad (approved):

omad:
Thanks Robbi, good catch, and good fix! 😀 

I've updated the STAC document in S3, and updated the ODC Index in AWS Dev and AWS Prod (with `s3-to-dc  --stac --update --allow-unsafe s3://dea-public-data/abares_clum_2020/\*.stac-item.json abares_clum_2020`).

PR #1243, opened 2024-03-14
Add DEA Intertidal microtidal style, update adaptive layer name

robbibt:
Adds a new microtidal elevation style, updates name of adaptive layer (currently not working)


erialC-P (approved):
Looks good to me!

PR #1244, opened 2024-03-15
Fix paths in ABARES STAC

robbibt:
We can't currently access ABARES data on dev or prod:
image

I think this is because some of the STAC paths are wrong. I've attempted to fix these below.

### DIFF
--- a/products/others/abares_clum_2020/abares_clum_2020.stac-item.json
+++ b/products/others/abares_clum_2020/abares_clum_2020.stac-item.json
@@ -374,7 +374,7 @@
       "type": "image/jpeg"
     },
     "alum_class": {
-      "href": "https://dea-public-data.s3-ap-southeast-2.amazonaws.com/projects/abares_clum_2020/clum_50m1220m.tiff",
+      "href": "clum_50m1220m.tiff",
       "type": "image/tiff; application=geotiff; profile=cloud-optimized",
       "title": "alum_class",
       "raster:bands": [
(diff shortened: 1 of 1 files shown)


products/others/abares_clum_2020/abares_clum_2020.stac-item.json
```diff
@@ -374,7 +374,7 @@
       "type": "image/jpeg"
     },
     "alum_class": {
-      "href": "https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/abares_clum_2020/clum_50m1220m.tiff",
+      "href": "https://dea-public-data.s3-ap-southeast-2.amazonaws.com/projects/abares_clum_2020/clum_50m1220m.tiff",
```
robbibt, line 377: Can this just be a relative link instead?

products/others/abares_clum_2020/abares_clum_2020.stac-item.json
```diff
@@ -374,7 +374,7 @@
       "type": "image/jpeg"
     },
     "alum_class": {
-      "href": "https://dea-public-data-dev.s3-ap-southeast-2.amazonaws.com/abares_clum_2020/clum_50m1220m.tiff",
+      "href": "clum_50m1220m.tiff",
```
robbibt, line 377: Is using a relative link here OK? The GeoTIFF should always live next to the metadata

robbibt:
Ah! Looks like both these changes were already made in the STAC metadata available on S3 here - just not copied back to this repo: https://data.dea.ga.gov.au/projects/abares_clum_2020/abares_clum_2020.stac-item.json

omad (approved):

omad:
Thanks Robbi. I did some fat fingering too with putting the STAC doc into the wrong path. 

Should be fixed in both environments now.

PR #1246, opened 2024-03-18
Add DEA Intertidal adaptive uncertainty layer

robbibt:
This adds a new adaptive uncertainty style to match the adaptive elevation style. It scales colours from uncertainty of 10% to 30% of the observed tide range.

Also makes some minor tweaks to the adaptive elevation style.


vnewey (approved):
Looks great!

PR #1247, opened 2024-03-19
Update styling, reorder styles, add abstract

robbibt:
This PR:

- Updates the lower colour limit in the adaptive uncertainty style to avoid completely black pixels
- Reorders the styles to put the adaptive style up to (and set to default)
- Adds a draft abstract


vnewey (approved):
Looks good

PR #1251, opened 2024-03-27
Remove extents from DEA Intertidal OWS style

robbibt:
Temporarily remove Extents style from DEA Intertidal OWS layer (data has been removed from the first release of the product)


SpacemanPaul (approved):

omad (approved):
Temporarily?

Also, does that `inventory.json` help anyone, or should it be gotten rid of?

PR #1253, opened 2024-04-04
Prod OWS DEA Intertidal layer

robbibt:
This PR adds DEA Intertidal to prod OWS and the prod section of Terriacube, including a simpler set of three styles (elevation, uncertainty and exposure).

It also updates the prod OWS and prod section of Terriacube to place the older NIDEM and ITEM layers inside "Other".


erialC-P (approved):
Awesome, looks good to go!

PR #1258, opened 2024-04-15
Added LS-9 to FC processing configuration.

tebadi:
Added configurations for enabling processing of FC on LS9.


emmaai (changes requested):
the correct config is this one
https://github.com/GeoscienceAustralia/dea-config/blob/master/prod/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml

PR #1260, opened 2024-04-17
Re-align dev and prod Alchemist config

matt-fry-ga:
Appears that changes to Alchemist config have been made in production and not development.  
The PR aligns the two sets of config such tht they use the same values, with the exception of the output location and explorer url.


dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
```diff
@@ -55,14 +65,14 @@ output:
   metadata:
     product_family: fc
     producer: ga.gov.au
-    dataset_version: 2.5.0
+    dataset_version: 2.5.1
     collection_number: 3
-    naming_conventions: dea_c3 
+    naming_conventions: dea_c3
   properties:
     dea:dataset_maturity: interim
```
tebadi, line 72: The dataset maturity should be final.

dev/services/alchemist/ga_ls_fc_3/ga_ls_fc_3.alchemist.yaml
```diff
@@ -2,7 +2,6 @@ specification:
   products:
```
tebadi, line 2: We can also delete LS8 coefficients under transform_args_per_product section given LS8 is processed using the dedicated config.

tebadi (approved):
The FC configs look good. I added Emma to check the wofs related changes.

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
emmaai, line 1: I think this yaml is particularly used for usgs collection 2 as indicated in line 63. DEAfrica used to need it. I'm not sure if it was changed since. Better just keep it, and make another copy of `ga_ls8_fc_3.yaml` in the folder. I'd say divergence between dev and prod is inevitable.

emmaai:
I'm not really picky about alignment between prod and dev. Usually we make changes in dev then promote them into prod. However, if there is no need for this process, then the divergence would happen. In that sense, every change in dev should by default have current prod as the start point.

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
matt-fry-ga: For dev?

dev/services/alchemist/ga_ls_fc_3/ga_ls_fc_3.alchemist.yaml
matt-fry-ga: is this for both dev and prod?  It is in both

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
matt-fry-ga: I will follow your suggestion @emmaai .  
Not that there is no equivalent file in prod

dev/services/alchemist/ga_ls_fc_3/ga_ls_fc_3.alchemist.yaml
tebadi: For both. There is nothing relying on this given this is lS5/7 configuration file, right @emmaai?

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
tebadi: This is slightly different. Alchemist is a streaming pipeline and we will run all the datasets in the production. I'll only use this config for the testing of LS9 before I run everything in the prod.

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
matt-fry-ga: I've made a copy of the original unaltered file as requested.

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
matt-fry-ga: @emmaai I think we should be testing in dev before releasing to prod. I don't think that has changed, just that people are maybe ignoring that process?

dev/services/alchemist/ga_ls_fc_3/ga_ls_fc_3.alchemist.yaml
matt-fry-ga: Done

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
matt-fry-ga: > I don't think it inherited the `maturity` from ard. The datasets can be labeled as interim particularly if they may require to be indexed in dev database. It's a good way to seperate them from "official" datasets. As mentioned, there is reason that dev diverged from prod. If we don't really wanna test different config options as a development process, might better just use prod, even we run pipeline on dev cluster.

I've changed the maturity to final, but I'm still uneasy about it

dev/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml
```diff
@@ -0,0 +1,68 @@
...
+          - 1.00446
+        nir:
+          - 0.4
+          - 0.98906
+        swir1:
+          - 25.6
+          - 0.99467
+        swir2:
+          - -32.7
+          - 1.02551
```
tebadi, line 38: we're still missing the output coefficients. Please see https://github.com/GeoscienceAustralia/dea-config/blob/master/prod/services/alchemist/ga_ls_fc_3/ga_ls8_fc_3.alchemist.yaml

matt-fry-ga: Done

tebadi:
> The FC configs look good. I added Emma to check the wofs related changes.

This looks good Mathew!

PR #1265, opened 2024-05-16
DEA Waterbodies v3.0.0 Terria Map update version numbering and wording consistency

LaurenSchenk1:
Update DEA Waterbodies prod Terria Map documentation for consistent version numbering (3.0.0) and consistent use of "timeseries" vs "time series" and "waterbodies" vs "water bodies".


BexDunn (approved):
looks good Lauren

PR #1266, opened 2024-05-23
Add expanded grid to Terria and DEA Maps

robbibt:
This PR adds a GeoJSON version of the updated/expanded Collection 3 grid to Terriacube and DEA Maps (see preview links below). It is located in `Other > DEA Shapes`.

To make it clear that the grid is not final and may change, I have included the following caveat:

> This is a provisional layer intended to replace the existing <b>DEA Summary Product Grid (Collection 3)</b> by updating region code IDs to account for new satellite data processed over Australia's external territories.

image

### DIFF
--- a/dev/terria/terria-cube-v8.json
+++ b/dev/terria/terria-cube-v8.json
@@ -1820,10 +1820,10 @@
                                     "info": [
                                         {
                                             "name": "Important information",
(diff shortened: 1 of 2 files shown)


gumpgourou (approved):
Maybe add "....to replace during 2024 the....  "?

PR #1279, opened 2024-08-14
Fix masking in remaining ARD styles

robbibt:
Have updated:

- Dev Sentinel-2 layer
- Prod Landsat layer
- Prod Sentinel-2 layer
- Prod combined layer

Dev WMS combined style is old and does not apply masking - we should fix this but not in this PR


robbibt:
@SpacemanPaul Should we remove the broken CFG Test actions from this repo? They haven't worked for months and only add noise...

SpacemanPaul (approved):

PR #1280, opened 2024-08-22
Test Coastlines 2023 on dev

robbibt:
This makes a small change to TerriaCube to allow plotting the new 2023 Coastlines data. In particular, enables a drop-down style selection menu:
image


vnewey (approved):
Looks great Robbi!

PR #1281, opened 2024-08-23
Update DEA Intertidal featureinfo results

robbibt:
Added download link to main DEA Intertidal popup:
image

Adds a nice FeatureInfo popup for the difference tool:
image


vnewey (approved):
Looks good Robbi!

PR #1282, opened 2024-08-26
Update DEA Coastlines 2023 on DEA Maps

robbibt:
Updates DEA Maps for 2023 coastlines update 
(proper styles handling and 2023 data on graph)


vnewey (approved):
Looks good.
Approved, just need to follow up with Hellfish regarding extent of the annual shorelines supplementary wms

PR #1288, opened 2024-09-18
Geomad decom dev 2

JM-GA:
These two configs appear to still be used by other web services, adding them back in for now


SpacemanPaul (approved):
Oops - I did try to check for that, but obviously didn't do it properly.

PR #1296, opened 2024-11-11
Land Cover Release

JM-GA:
Product definitions, odc-stats configs and ows to test in dev


SpacemanPaul (approved):
All makes sense to me.   Let's throw it against the wall and see what sticks.

JM-GA:
Thanks @SpacemanPaul  - will deploy and see what happens

PR #1332, opened 2025-04-29
Add test of DEA Tidal Composites web services

robbibt:
This PR adds a test of DEA Tidal Composites dev web services, just with a small set of styles for now (low and high tide true colour, false colour, MNDWI, and clear count).


robbibt:
Hey @SpacemanPaul, any chance you could have a quick glance at this PR? It passes tests and I've tried out the styles locally.

PR #1333, opened 2025-04-29
Nbic uplift

biancafichera:
adding styling for new NBIC burnt area products to test


Kooie-cate (approved):
Well done, I'm looking forward to seeing it on dea maps :)

PR #1340, opened 2025-05-01
Hazards updates

Kooie-cate:
Adding Sentinel 2b to FMC 
fixing styling to NBIC stacked area
removed depreciated burnt area mapping layers from Terria (not from dev WMS)


biancafichera (approved):
Such a hard worker!

PR #1344, opened 2025-05-06
Experimental tidal composite OWS updates

robbibt:
This PR adds low and high tide threshold layers to OWS so we can hopefully use them in a FeatureInfo popup.

It also adds two experimental log-scaled true colour styles, intended to test whether we can use a single style that works well in dark and bright coastal environments (e.g. avoiding bright beaches from being washed out).


vnewey (approved):
looks good!

PR #1346, opened 2025-05-07
Test setting default time on tidal composites, update styling

robbibt:
Set DEA Tidal Composites to default to 2022 data

Use a softer transform (log instead of log10) in experimental layer to improve visualisation


erialC-P (approved):
Good to go!

PR #1350, opened 2025-05-12
Add Extents layer to DEA Intertidal, fix tide graph

robbibt:
Adds Extents style
Fixes broken tide graph links and metadata
image


erialC-P (approved):
approved - good catch!

PR #1358, opened 2025-06-12
correct fmc mosaic

Kooie-cate:


LaurenSchenk1:
Suggestion for future: in title and abstract capitalise A, B, C but otherwise looks good!

PR #1370, opened 2025-08-12
Add minimal OWS configuration for Sentinel-1 normalised radar backscatter

caitlinadams:
Minimal OWS Config including:
* layer config
* bands config
* style config

### DIFF
--- a/dev/services/wms/ows_refactored/baseline_satellite_data/sentinel1/style_s1_nrb_cfg.py
+++ b/dev/services/wms/ows_refactored/baseline_satellite_data/sentinel1/style_s1_nrb_cfg.py
@@ -2,10 +2,11 @@ style_s1_nrb_vv_gamma0 = {
     "name": "vv_gamma0",
     "title": "Linear backscatter VV gamma0",
     "components": {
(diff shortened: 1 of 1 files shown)


dev/services/wms/ows_refactored/baseline_satellite_data/sentinel1/style_s1_nrb_cfg.py
```diff
@@ -0,0 +1,11 @@
+style_s1_nrb_vv_gamma0 = {
+    "name": "vv_gamma0",
+    "title": "Linear backscatter VV gamma0",
+    "components": {
+        "red": {"VV_gamma0": 1},
+        "green": {"VV_gamma0": 1},
+        "blue": {"VV_gamma0": 1},
+    },
+}
```
SpacemanPaul, line 9: You will need to set a scale range.  See https://datacube-ows.readthedocs.io/en/latest/cfg_component_styles.html#component-scale-range

SpacemanPaul (changes requested):
Almost there - still a tiny bit *too* minimal.

SpacemanPaul (approved):

PR #1422, opened 2026-04-16
Update coastal product metadata to latest version from 2026 annual product updates

robbibt:
Saving the product definitions we used for the 2026 coastal product updates here for future reference (see "Service Request: Index coastal products into the prod datacube database using STAC metadata"). 

We have switched to the default `eo3` metadata type in the latest 2.1.0 and 1.1.0 versions of Intertidal and Tidal Composites.


erialC-P (approved):
Saving these is a good idea! Approved

PR #1424, opened 2026-04-20
Update DEA Intertidal and Tidal Composites in prod OWS for annual product release

robbibt:
## Approved in CAB 20260414.1; do not merge until datacube-ows has been upgraded to 1.9.12 in prod

This PR makes three minor changes to the DEA Intertidal and Tidal Composites products for the annual product update:

* DEA Intertidal and DEA Tidal Composites are now provided as continental COGs which include high level overviews for visualising data when zoomed out to continental scale. This PR switches the prod OWS layers to use unlimited resource limits (`reslim_wms_unlimited`), which enables continental previews when zoomed out (when previously only a blue box would be seen).
* Tide graph paths are now generated using a new approach that works on our continental COG data, which now generates file paths based on the coordinate of the GetFeatureInfo request rather than the tile region code (as we no longer index our tiled data).
* Minor updates to DEA Maps FeatureInfoTemplate pop-ups to simplify wording, and remove the 32km coastal grid now that it is no longer needed for continental COG data access.

These changes have been tested in dev and are working very well. They only affect DEA Intertidal and DEA Tidal Composite layers in prod OWS.

image

image


prod/services/wms/ows_refactored/sea_ocean_coast/tidal_composites_c3/utils_tidal_composites.py
```diff
@@ -9,21 +9,3 @@ def log_scaling(data, band):
     attractive three-band image.
     """
     return np.log(data[band] + 1)
-
-
-def tide_graph_path(data, ds):
```
robbibt, line 14: For context: removing this duplicated function here and re-using the one at `ows_refactored.sea_ocean_coast.intertidal_c3.utils_intertidal.tide_graph_path` instead to reduce future maintenance

prod/terria/dea-maps-v8.json
```diff
@@ -1187,20 +1187,11 @@
...
                                     },
                                     "tileWidth": 512,
                                     "tileHeight": 512,
                                     "id": "ghtu564"
-                                },
-                                {
-                                    "type": "geojson",
```
robbibt, line 1197: For context: removing this 32 km coastal processing grid from the DEA Maps catalogue as it is no longer relevant for our continental COG layers. It is still retained in Terriacube for future troubeshooting purposes.

SpacemanPaul (approved):

PR #1435, opened 2026-07-16
Update DEA Maps grid URLs to prod

ankurbhardwa:
Updates DEA Maps config to point the Sentinel-2 MGRS grid and Landsat path/row grid to prod URLs instead of dea-public-data-dev.

Updated URLs:
- https://data.dea.ga.gov.au/derivative/ga_s2_mgrs_tile_grid.geojson
- https://data.dea.ga.gov.au/derivative/ga_ls_path_row_grid.geojson

Validation:
- Both prod URLs return HTTP/2 200
- Change is limited to two URL updates in prod/terria/dea-maps-v8.json

Note:
- Existing dea-public-data-dev entry in corsDomains was left unchanged, as it is not a dataset URL.

### DIFF
--- a/prod/terria/dea-maps-v8.json
+++ b/prod/terria/dea-maps-v8.json
@@ -1589,13 +1589,13 @@
                                 {
                                     "type": "geojson",
                                     "name": "Sentinel-2 MGRS Tile grid for Australia",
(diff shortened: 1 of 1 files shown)


robbibt (approved):
Looks great, thanks @ankurbhardwa!

